PR #16, opened 2021-01-26
Add unit tests for azure-quantum

nextstate:


azure-quantum/azure/quantum/job.py
```diff
@@ -82,3 +83,9 @@ def get_results(self):
 
         result = json.loads(payload.decode('utf8'))
         return result
+            
+    @staticmethod
+    def create_job_id() -> str:
+        """Create a unique id for a new job.
+        """
+        return uuid.uuid1()
```
guenp, line 91: Strictly speaking, `uuid.uuid()` returns a `UUID` object. If you want this to return a string as the type hint suggests I would change this to 
```suggestion
        return str(uuid.uuid1())
```

azure-quantum/tests/unit/recording_updater.py
```diff
@@ -0,0 +1,61 @@
...
+class RecordingUpdater:
+    """RecordingUpdater
+    
+    Replaces sensitive data with dummy values in network recordings.
+    """
+    dummy_uid = "00000000-0000-0000-0000-000000000000"
+    dummy_rg = "dummy-rg"
+    dummy_ws = "dummy-ws"
+
+    def __init__(self, recording_directory, recording_glob):
```
guenp, line 26: Docstrings and type hints would be useful for these methods, it's not strictly required for internal methods though but for the constructor and `update_recordings_with_dummy_values` would be helpful.

azure-quantum/tests/unit/recording_updater.py
```diff
@@ -0,0 +1,61 @@
...
+
+import glob
+import os
+import re
+
+
+class RecordingUpdater:
+    """RecordingUpdater
+    
+    Replaces sensitive data with dummy values in network recordings.
```
guenp, line 20:
```suggestion
    Replaces secrets with dummy values in network recordings.
```

azure-quantum/tests/unit/test_job.py
```diff
@@ -0,0 +1,111 @@
...
+            'version': '1.0',
+            'configuration': {'0': 1, '1': 1, '2': -1, '3': 1, '4': -1},
+            'cost': -6.0,
+            'parameters': {'beta_start': 0.2, 'beta_stop': 1.9307236000000003, 'restarts': 360, 'sweeps': 50}}
+
+        self.assertEqual(expected, actual)
+        
+
+def update_recordings_with_dummy_values():
+    recording_directory = os.path.abspath(os.path.join(os.path.dirname(__file__), "recordings"))
```
guenp, line 102: I would add something like this for posterity to make it clear this is not a unit test:
```suggestion
    """Replace all secrets in the recorded .yaml files with dummy values as defined in RecordingUpdater. This only needs to be run once."""
    recording_directory = os.path.abspath(os.path.join(os.path.dirname(__file__), "recordings"))
```

azure-quantum/tests/unit/workspace_init.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+import configparser
+import unittest.mock
+
+from azure.quantum import Workspace
+from azure.quantum.workspace import MsalWrapper
+
+from recording_updater import RecordingUpdater
+
+def create_workspace_mock_login(**kwds) -> Workspace:
```
guenp, line 20:
```suggestion
def create_workspace_mock_login(**kwds) -> Workspace:
"""Create a mock Workspace object by patching the Azure authentication and uses a dummy authentication token."""
```

azure-quantum/tests/unit/test_workspace.py
```diff
@@ -14,6 +14,18 @@
...
 from azure.quantum import Workspace
 
+from workspace_init import create_workspace_mock_login
+
+def test_workspace_login():
+    subscription_id = "44ef49ad-64e4-44e5-a3ba-1ee87e19d3f4"
+    resource_group = "rg"
+    name = "n"
+
+    ws = create_workspace_mock_login(
```
guenp, line 24: Just a suggestion to make it clear that we covered this method in test
```suggestion
    # This tests Workspace.login() using a mock authentication token
    ws = create_workspace_mock_login(
```

guenp (approved):
Looks great, just have some minor comments/suggestions. Thanks!

Issue #22, opened 2021-02-17
Service Principal auth broken in latest azure-quantum


Mobius5150:
Using the latest version of the azure-quantum package and following the [official service principal auth instructions](https://docs.microsoft.com/en-us/azure/quantum/optimization-authenticate-service-principal) shows that auth is currently broken. Attempting to authenticate yields the following exception:

```
  File "\Python\Python36\lib\site-packages\msrest\pipeline\__init__.py", line 197, in run
    return first_node.send(pipeline_request, **kwargs)  # type: ignore
  File "\Python\Python36\lib\site-packages\msrest\pipeline\__init__.py", line 150, in send
    response = self.next.send(request, **kwargs)
  File "\Python\Python36\lib\site-packages\msrest\pipeline\requests.py", line 68, in send
    request.context.session = session = self._creds.signed_session()
  File "\Python\Python36\lib\site-packages\msrest\authentication.py", line 118, in signed_session
    header = "{} {}".format(self.scheme, self.token['access_token'])
TypeError: 'ServicePrincipalCredentials' object is not subscriptable
```

It seems that this is also using an old pattern, and that Azure SDKs are moving to the `azure-identity` package instead of `azure-common`.

ashwinmayya:
Please using the following command to revert to an older version of the package to unblock this scenario while we fix the issue.
pip install azure-quantum==0.15.2101.125897 --force-reinstall

PR #23, opened 2021-02-17
Add Streaming Problem Interface

Mobius5150:
**Summary**
Adds a new `StreamingProblem` class, which allows users to construct a problem without having to hold the whole problem definition in memory at one time, which may be a significant problem when constructing large problems or when constructing many problems in parallel.

Sample Usage:
```python
import StreamingProblem from azure.quantum.optimization
workspace = Workspace(...)
problem = StreamingProblem(workspace = workspace, name="My Problem", problem_type=ProblemType.ising)
for terms in generate_problem_terms():
    problem.add_terms(terms)
results = solver.optimize(problem)
```

Under the hood, the `StreamingProblem` uploads the problem in chunks as terms are added using [Block Blobs](https://docs.microsoft.com/rest/api/storageservices/understanding-block-blobs--append-blobs--and-page-blobs). The problem representation is committed in Azure storage when the user calls problem.upload(). If the user never finalizes the problem by submitting it or calling `problem.upload()` Azure will automatically clean up the uncommitted blocks.

StreamingProblem is otherwise used in the exact same way as the existing `Problem` class, with the exception that you cannot read back terms after they have been added. However, after you have called `upload` you may call `download` to download the complete problem definition back to  the client.

**Open Issues**
This is a somewhat large PR - very open to feedback on the approach and structure!

I've added integration tests for `StreamingProblem`, however the format is using the unittest framework which differs from the existing integration tests. I also encountered #22 which, once resolved, will allow us to automate integration tests in builds. I'd appreciate maintainer feedback on how these are structured.

### DIFF
--- a/azure-quantum/azure/quantum/optimization/problem.py
+++ b/azure-quantum/azure/quantum/optimization/problem.py
@@ -55,6 +55,7 @@ class Problem:
         self.uploaded_blob_params = None
 
     def serialize(self) -> str:
+        """Serializes the problem to a JSON string"""
         result = {
             "cost_function": {
                 "version": "1.1" if self.init_config else "1.0",
@@ -69,7 +70,14 @@ class Problem:
         return json.dumps(result)
 
     @staticmethod
-    def deserialize(string, name):
+    def deserialize(string: str, name: str):
+        """Deserializes the problem from a JSON string
+
+        :param string: The string to be deserializes to a `Problem` instance
+        :type string: str
+        :param name: The name of the problem
+        :type name: str
+        """
         result = json.loads(string)
         problem = Problem(
             name = name,
@@ -83,10 +91,21 @@ class Problem:
         return problem
  
     def add_term(self, c: Union[int, float], indices: List[int]): 
+        """Adds a single term to the `Problem` representation
+
+        :param c: The cost or weight of this term
+        :type c: int, float
+        :param indices: The variable indices that are in this term
+        :type indices: List[int]
+        """
         self.terms.append(Term(indices=indices, c=c))
         self.uploaded_blob_uri = None
     
     def add_terms(self, terms: List[Term]): 
+        """Adds a list of terms to the `Problem` representation
+
+        :param terms: The list of terms to add to the problem
+        """
         self.terms += terms
         self.uploaded_blob_uri = None
 
--- a/azure-quantum/azure/quantum/optimization/problem.py
+++ b/azure-quantum/azure/quantum/optimization/problem.py
@@ -153,11 +153,11 @@ class Problem:
             # No storage account is passed, use the linked one
             container_uri = workspace._get_linked_storage_sas_uri(container_name)
             container_client = ContainerClient.from_container_url(container_uri)
-            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), returnSasToken=False)
+            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), return_sas_token=False)
         else:
             # Use the specified storage account
             container_client = ContainerClient.from_connection_string(workspace.storage, container_name)
-            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), returnSasToken=True)
+            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), return_sas_token=True)
 
         self.uploaded_blob_params = blob_params
         return self.uploaded_blob_uri
--- a/azure-quantum/azure/quantum/optimization/problem.py
+++ b/azure-quantum/azure/quantum/optimization/problem.py
@@ -48,9 +48,11 @@ class Problem:
         problem_type: ProblemType = ProblemType.ising
     ):
         self.name = name
-        self.terms = terms if terms is not None else []
+        self.terms = terms.copy() if terms is not None else []
         self.problem_type = problem_type
         self.init_config = init_config
+        self.uploaded_blob_uri = None
+        self.uploaded_blob_params = None
 
     def serialize(self) -> str:
         result = {
@@ -66,11 +68,27 @@ class Problem:
 
         return json.dumps(result)
 
-    def add_term(self, c: Union[int, float], indices: List[int]):
+    @staticmethod
+    def deserialize(string, name):
+        result = json.loads(string)
+        problem = Problem(
+            name = name,
+            terms = [Term.from_dict(t) for t in result['cost_function']['terms']],
+            problem_type = ProblemType[result['cost_function']['type']]
+        )
+        
+        if 'initial_configuration' in result['cost_function']:
+            problem.init_config = result['cost_function']['initial_configuration']
+
+        return problem
+ 
+    def add_term(self, c: Union[int, float], indices: List[int]): 
         self.terms.append(Term(indices=indices, c=c))
+        self.uploaded_blob_uri = None
     
-    def add_terms(self, terms: List[Term]):
+    def add_terms(self, terms: List[Term]): 
         self.terms += terms
+        self.uploaded_blob_uri = None
 
     def upload(
         self,
@@ -92,6 +110,10 @@ class Problem:
         :return: uri of the uploaded problem
         :rtype: [type]
         """
+        blob_params = [workspace, container_name, blob_name, compress]
+        if self.uploaded_blob_uri and self.uploaded_blob_params == blob_params:
+            return self.uploaded_blob_uri
+        
         if blob_name is None:
             blob_name = '{}-{}'.format(self.name, uuid.uuid1())
             
@@ -112,8 +134,11 @@ class Problem:
             # No storage account is passed, use the linked one
             container_uri = workspace._get_linked_storage_sas_uri(container_name)
             container_client = ContainerClient.from_container_url(container_uri)
-            return upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), return_sas_token=False)
+            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), returnSasToken=False)
         else:
             # Use the specified storage account
             container_client = ContainerClient.from_connection_string(workspace.storage, container_name)
-            return upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), return_sas_token=True)
+            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), returnSasToken=True)
+
+        self.uploaded_blob_params = blob_params
+        return self.uploaded_blob_uri
--- a/azure-quantum/azure/quantum/optimization/problem.py
+++ b/azure-quantum/azure/quantum/optimization/problem.py
@@ -80,9 +80,9 @@ class Problem:
         """
         result = json.loads(string)
         problem = Problem(
-            name = name,
-            terms = [Term.from_dict(t) for t in result['cost_function']['terms']],
-            problem_type = ProblemType[result['cost_function']['type']]
+            name=name,
+            terms=[Term.from_dict(t) for t in result['cost_function']['terms']],
+            problem_type=ProblemType[result['cost_function']['type']]
         )
         
         if 'initial_configuration' in result['cost_function']:
--- a/azure-quantum/azure/quantum/optimization/problem.py
+++ b/azure-quantum/azure/quantum/optimization/problem.py
@@ -69,16 +69,16 @@ class Problem:
 
         return json.dumps(result)
 
-    @staticmethod
-    def deserialize(string: str, name: str):
-        """Deserializes the problem from a JSON string
+    @classmethod
+    def deserialize(cls, problem_as_json: str, name: str):
+        """Deserializes the problem from a JSON string serialized with Problem.serialize()
 
-        :param string: The string to be deserializes to a `Problem` instance
-        :type string: str
+        :param problem_as_json: The string to be deserialized to a `Problem` instance
+        :type problem_as_json: str
         :param name: The name of the problem
         :type name: str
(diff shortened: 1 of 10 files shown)


azure-quantum/tests/integration/test_streaming_problem.py
```diff
@@ -0,0 +1,109 @@
...
+        size_thresh: int,
+        compress: bool,
+        problem_type: ProblemType = ProblemType.ising,
+        initial_terms: List[Term] = [],
+        **kwargs
+    ):
+        ws = _init_ws_()
+        sProblem = StreamingProblem(
+            ws,
+            name = "test",
```
guenp, line 56:
```suggestion
            name="test",
```

guenp (commented):
Thanks for adding this feature! Before I do a more thorough review, would you mind adding docstrings to the public methods you've added?

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -66,11 +68,27 @@ def serialize(self) -> str:
 
         return json.dumps(result)
 
-    def add_term(self, c: Union[int, float], indices: List[int]):
+    @staticmethod
+    def deserialize(string, name):
```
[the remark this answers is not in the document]
Mobius5150: Added!

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -66,11 +68,27 @@ def serialize(self) -> str:
...
+        result = json.loads(string)
+        problem = Problem(
+            name = name,
+            terms = [Term.from_dict(t) for t in result['cost_function']['terms']],
+            problem_type = ProblemType[result['cost_function']['type']]
```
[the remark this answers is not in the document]
Mobius5150: Will apply these changes - is there a format guide that the repo follows that I may check the change against myself?

Mobius5150:
> 
> 
> Thanks for adding this feature! Before I do a more thorough review, would you mind adding docstrings to the public methods you've added? Especially `add_terms` as the behavior of this method is now very different from the other base class in its interaction with the service.

I've added docstrings, but the `add_term(s)` methods behave relatively consistently with the `Problem` class - they add one or more terms to the problem representation. The _way_ in which terms are added is I think an implementation detail that is documented on the `StreamingProblem` class itself. 

I guess what I'd like to convey is that I see `Problem` as an interface for which there may be multiple implementations. Direct inheritance isn't an option but I could instead add a base class that this and `Problem` both inherit from if that's clearer?

guenp:
@Mobius5150 From what I understand `StreamingProblem.add_terms` already starts uploading the data, while the regular `Problem` classes don't. Is that correct?

azure-quantum/azure/quantum/storage.py
```diff
@@ -53,14 +54,7 @@ def get_container_uri(connection_string: str, container_name: str) -> str:
...
 
-def upload_blob(
-    container: ContainerClient,
-    blob_name: str,
-    content_type:str,
-    content_encoding:str,
-    data: Any,
-    return_sas_token: bool=True
-) -> str:
+def upload_blob(container: ContainerClient, blob_name: str, content_type:str, content_encoding:str, data: Any, returnSasToken: bool=True) -> str:
```
guenp, line 57: Why use camel case for `returnSasToken`? [PEP8 ](https://pep8.org/)recommends  using snake case.

azure-quantum/azure/quantum/storage.py
```diff
@@ -74,24 +68,57 @@ def upload_blob(
...
+        uri = get_blob_uri_with_sas_token(blob)
+    else:
+        uri = remove_sas_token(blob.url)
 
-        uri = blob.url + "?" + sas_token
+    logger.debug(f"  - blob access url: '{uri}'.")
+
+    return uri
+
+def append_blob(container: ContainerClient, blob_name: str, content_type:str, content_encoding:str, data: Any, returnSasToken: bool=True, metadata: Dict[str, str] = None) -> str:
```
guenp, line 80: This line is a bit long, to shorten it I'd suggest adding newlines after each comma

azure-quantum/azure/quantum/storage.py
```diff
@@ -74,24 +68,57 @@ def upload_blob(
...
+        uri = get_blob_uri_with_sas_token(blob)
     else:
         uri = remove_sas_token(blob.url)
 
     logger.debug(f"  - blob access url: '{uri}'.")
 
     return uri
 
+def get_blob_uri_with_sas_token(blob: BlobClient):
+    sas_token = generate_blob_sas(
```
guenp, line 111: Missing docstring

azure-quantum/azure/quantum/storage.py
```diff
@@ -104,10 +131,109 @@ def download_blob(blob_url: str) -> Any:
 
     return response
 
+def download_blob_properties(blob_url: str) -> Dict[str, str]:
```
guenp, line 134: Same here, looks like there are some missing docstrings

azure-quantum/azure/quantum/storage.py
```diff
@@ -104,10 +131,109 @@ def download_blob(blob_url: str) -> Any:
...
+
 def remove_sas_token(sas_uri: str) -> str:
     index = sas_uri.find('?')
     if index != -1:
         sas_uri = sas_uri[0:index]
     
     return sas_uri
 
+
+def init_blob_for_streaming_upload(container: ContainerClient, blob_name: str, content_type:str, content_encoding:str, data: Any, returnSasToken: bool=True) -> str:
```
guenp, line 159: This line is a bit long, could you shorten it?

guenp: Same comment here on camel case vs. snake case, `returnSasToken` should be `return_sas_token`

azure-quantum/azure/quantum/storage.py
```diff
@@ -104,10 +131,109 @@ def download_blob(blob_url: str) -> Any:
...
+class StreamedBlob:
+    def __init__(self, container: ContainerClient, blob_name: str, content_type: str, content_encoding:str):
+        self.container = container
+        self.blob_name = blob_name
+        self.content_settings = ContentSettings(content_type=content_type, content_encoding=content_encoding)
+        self.state = StreamedBlobState.not_initialized
+        self.blob = container.get_blob_client(blob_name)
+        self.blocks = []
+
+    def upload_data(self, data):
```
guenp, line 207: Would you mind adding some docstrings to the methods here?

azure-quantum/azure/quantum/optimization/streaming_problem.py
```diff
@@ -0,0 +1,322 @@
...
+            container_uri = self.workspace._get_linked_storage_sas_uri(self.id)
+            container_client = ContainerClient.from_container_url(container_uri)
+        else:
+            # Use the specified storage account
+            container_client = ContainerClient.from_connection_string(self.workspace.storage, self.id)
+        
+        return { 'blob_name': blob_name, 'container_client': container_client }
+
+    def add_terms(self, terms: List[Term]) :
+        """Adds a list of terms to the `Problem` representation
```
guenp, line 107: Would you add a note here about what's happening behind the scenes with uploading the terms?

guenp (commented):
Hi @Mobius5150 , I just have some comments and clarifying questions.

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -48,11 +48,14 @@ def __init__(
         problem_type: ProblemType = ProblemType.ising
     ):
         self.name = name
-        self.terms = terms if terms is not None else []
+        self.terms = terms.copy() if terms is not None else []
```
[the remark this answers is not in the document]
Mobius5150: It's copied in the constructor because the add_term/add_terms methods append terms to the list - modifying the same refence of the list that you passed in which can cause unexpected behavior for the user who may expect the array that they passed in to not have been modified (I encountered this).

azure-quantum/azure/quantum/storage.py
Mobius5150: Form of habit from writing Typescript :)

Mobius5150:
@guenp thanks for the feedback! I've updated the PR

azure-quantum/azure/quantum/storage.py
guenp, line 57:
```suggestion
def upload_blob(container: ContainerClient, blob_name: str, content_type:str, content_encoding:str, data: Any, return_sas_token: bool=True) -> str:
```

azure-quantum/azure/quantum/storage.py
```diff
@@ -104,10 +140,146 @@ def download_blob(blob_url: str) -> Any:
...
+            raise Exception('StreamedBlob cannot commit before uploading data')
+        elif self.state == StreamedBlobState.committed:
+            raise Exception('StreamedBlob is already committed')
+
+        logger.debug(f"Committing {len(self.blocks)} blocks {self.blob_name}")
+        self.blob.commit_block_list(self.blocks, content_settings=self.content_settings, metadata=metadata)
+        self.state = StreamedBlobState.committed
+        logger.debug(f"Committed {self.blob_name}")
+
+    def getUri(self, withSasToken : bool = False):
```
guenp, line 275:
```suggestion
    def getUri(self, with_sas_token: bool = False):
```

guenp, line 275:
```suggestion
    def getUri(self, with_sas_token: bool = False):
```

azure-quantum/azure/quantum/storage.py
```diff
@@ -104,10 +140,146 @@ def download_blob(blob_url: str) -> Any:
...
+        """Gets the full Azure Storage URI for the uploaded blob after it has been committed"""
+        if self.state != StreamedBlobState.committed:
+            raise Exception('Can only retrieve sas token for committed blob')
+        if withSasToken:
```
guenp, line 279:
```suggestion
        if with_sas_token:
```

azure-quantum/azure/quantum/storage.py
```diff
@@ -74,23 +68,65 @@ def upload_blob(
     blob.upload_blob(data, content_settings=content_settings)
     logger.debug(f"  - blob '{blob_name}' uploaded. generating sas token.")
 
+    if returnSasToken:
```
guenp, line 71:
```suggestion
    if return_sas_token:
```

guenp (approved):
Approved with some suggested changes :) Thanks @Mobius5150 !

guenp:
@vxfield would you mind taking a quick look at this PR before we merge? Thank you!

azure-quantum/tests/integration/test_streaming_problem.py
```diff
@@ -0,0 +1,109 @@
...
+
+from azure.quantum import Workspace
+from azure.quantum.optimization import StreamingProblem, Problem, ProblemType, Term
+from azure.quantum.storage import download_blob
+from azure.common.credentials import ServicePrincipalCredentials
+from azure.identity import DefaultAzureCredential
+
+def _init_ws_():
+    workspace = Workspace(
+        subscription_id=os.environ.get('TEST_WORKSPACE_SUB'),
```
vxfield, line 26: The following environment variables should be used since they are already standardized, instead of defining new env vars like `TEST_WORKSPACE_SUB`.

    client_id = os.environ.get("AZURE_CLIENT_ID","")
    client_secret = os.environ.get("AZURE_CLIENT_SECRET","")
    tenant_id = os.environ.get("AZURE_TENANT_ID","")
    resource_group = os.environ.get("RESOURCE_GROUP","")
    subscription_id = os.environ.get("SUBSCRIPTION_ID","")
    workspace_name = os.environ.get("WORKSPACE_NAME","")

vxfield (changes requested):
Thanks for this new feature, Mike.
Code looks good to me, and Guen already captured everything that I could think of except about the tests, which infrastructure has changed a bit since you forked your branch.

For example, notice that this [file](https://github.com/microsoft/qdk-python/blob/902004cb1f965a0830adb1c0f4a9e8125eb02648/azure-quantum/tests/integration/test_integration.py#L40 
) has changed to used environment variables instead of a config.ini file.

Could you merge the latest `main` branch into your branch?

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -112,8 +153,11 @@ def upload(
             # No storage account is passed, use the linked one
             container_uri = workspace._get_linked_storage_sas_uri(container_name)
             container_client = ContainerClient.from_container_url(container_uri)
-            return upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), return_sas_token=False)
+            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), returnSasToken=False)
```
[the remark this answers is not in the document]
vxfield: Looks like you forgot to update the parameter name here

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -112,8 +153,11 @@ def upload(
...
         else:
             # Use the specified storage account
             container_client = ContainerClient.from_connection_string(workspace.storage, container_name)
-            return upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), return_sas_token=True)
+            self.uploaded_blob_uri = upload_blob(container_client, blob_name, content_type, encoding, data.getvalue(), returnSasToken=True)
```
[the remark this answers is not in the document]
vxfield: Looks like you forgot to update the parameter name here

azure-quantum/tests/integration/test_streaming_problem.py
Mobius5150: Just pushed this change - @vxfield please let me know if this looks good now

vxfield (approved):
I was able to run the tests on my machine and they are working fine.
We will need to update the unit and integration tests, but that can be done later in April along other updates on the testing infrastructure.

PR #24, opened 2021-02-18
Add support for NumPy types in Term construction.

ricardo-espinoza:


azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -4,23 +4,45 @@
 ##
 
 from typing import List, Union, Optional
+import numpy as np
```
Mobius5150, line 7: Does this add numpy as a dependency?

ricardo-espinoza: Yes, after following up with you offline, we're revisiting the approach for this change.

cgranade: That can be mitigated by guarding with an `if typing.TYPE_CHECKING` statement: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -4,23 +4,45 @@
...
 
 __all__ = ['Term']
 
+numpy_integer_types = [np.byte, np.ubyte, np.short, np.ushort, np.intc, np.uintc,
+                       np.int_, np.uint, np.longlong, np.ulonglong, np.int8, np.uint8,
+                       np.int16, np.uint16, np.int32, np.uint32, np.int64, np.uint64]
+
+numpy_float_types = [np.float16, np.float32, np.float64, np.float_, np.half, np.single, np.double]
+
+def _convert_if_numpy_type(param : npt.DTypeLike):
```
cgranade, line 18: I suspect you want ArrayLike and not DTypeLike here? DTypeLike matches objects representing dtypes, and not values of those dtypes.

azure-quantum/azure/quantum/optimization/term.py
ricardo-espinoza: Per discussion offline, we're leaving the numpy dependency in, however, we check for numpy.typing conditionally to support this logic.

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -4,23 +4,45 @@
...
+    # Attempt first a conversion to a supported type if parameter is a numpy float/int.
+    if (type(param) in numpy_integer_types + numpy_float_types):
```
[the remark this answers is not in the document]
ricardo-espinoza: My understanding is that we would only expect a value in this case, however, changing to npt.ArrayLike to follow the rest of the code.

azure-quantum/azure/quantum/optimization/term.py
ricardo-espinoza: Thanks!

azure-quantum/azure/quantum/optimization/term.py
cgranade: Fair enough! NumPy is _extremely_ common in scientific Python audiences (may be nearly universal, even), but perhaps less so outside of scientific usage. If we believe that the primary users are likely to be scientific in nature, taking a NumPy dependency is nearly zero-weight as a result.

azure-quantum/azure/quantum/optimization/term.py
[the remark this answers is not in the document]
cgranade: May be able to resolve by looking at `param.dtype` in the case that `hasattr(param, 'dtype')`.

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -9,20 +9,52 @@
...
+
+    def _convert_if_numpy_type(param : WArray):
+        # Attempt first a conversion to a supported type if parameter is a numpy float/int.
+        numpy_integer_types = [np.byte, np.ubyte, np.short, np.ushort, np.intc, np.uintc,
+                               np.int_, np.uint, np.longlong, np.ulonglong, np.int8, np.uint8,
+                               np.int16, np.uint16, np.int32, np.uint32, np.int64, np.uint64]
+
+        numpy_float_types = [np.float16, np.float32, np.float64, np.float_, np.half, np.single, np.double]
+
+        if (type(param) in numpy_integer_types + numpy_float_types):
```
cgranade, line 24:
```suggestion
        if (hasattr(param, 'dtype') and param.dtype in numpy_integer_types + numpy_float_types):
```

cgranade, line 24: I'd also suggest dropping the parens around the conditional here, as per PEP 8 and/or black style guides.

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -9,20 +9,52 @@
...
 
+        if(type(w) != None and type(c) != None):
+            raise RuntimeError("Cost has been specified multiple times. Please do not specify 'w' if using 'c'.")
+
         if(w != None):
+            # Legacy support if 'w' is used to specify term instead of the expected 'c'.
+            self.w = _convert_if_numpy_type(w)
             if(type(w) != int and type(w) != float):
-                raise RuntimeError("w must be a float or int value.")
+                raise RuntimeError("w must be a float or int value, or a numpy value that can be converted to those.")
```
cgranade, line 50:
```suggestion
                raise RuntimeError("w must be a float or int value, or a NumPy value that can be converted to those.")
```

azure-quantum/azure/quantum/optimization/term.py
cgranade, line 24: Third suggestion on this line, but this condition doesn't check if an array is scalar-like (shape `()` or `(1,)`) before calling `.item()`. You may be able to give a more actionable error message by doing so.

ricardo-espinoza: Adding handling.

cgranade (approved):

PR #26, opened 2021-02-19
Require an explicit location during workspace specification.

ricardo-espinoza:
This change aligns the QDK Python requirement on workspace location as part of the API, instead of defaulting to a region which might not correspond with the provided workspace.


vxfield:
I had to do a big refactor (#54) to move to the auto-generated client to track2 so I incorporated this check there as well.
Once that PR merges, likely today, we can close this PR.

vxfield:
Closed as #54 already implemented this.

PR #33, opened 2021-03-17
add toshiba solver and tests

sanjgupt:


vxfield (approved):
Looks good to me.
We need to improve our automated tests and create unit tests for all the solvers (instead of just having the "manual" integration test scripts), but we can work on that on another PR.

PR #37, opened 2021-03-30
Fix build pipeline on release branch

vxfield:
(This is the same as PR #36 but applied to the release branch)

This PR fixes the build pipeline that stopped working due to a new integration test for the azure-quantum package (introduced with #23) and the inclusion of running tests as part of the main build (introduced with #20).

PR #23 added some integration unit tests in the azure-quantum, but we don't have the infra yet to run them from ADO, so I disabled the unit tests for now.
For now, one needs to manually set some environment variables described in the integration tests README.md and manually run those tests.
In April the "integration tests" and "unit tests" will be consolidated and will be able to run against a live-service and also in playback mode.

### DIFF
--- a/azure-quantum/environment.yml
+++ b/azure-quantum/environment.yml
@@ -13,3 +13,4 @@ dependencies:
     - msrestazure
     - msal
     - azure-common
+    - azure-identity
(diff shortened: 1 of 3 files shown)


bettinaheim (approved):
I am not too familiar with the latest changes to the test infrastructure, but it looks reasonable to me, reading through the description in the PR.

PR #38, opened 2021-03-30
updating tabu changes to parity

adelebai:
Updating tabu changes to parity.
These were checked into the old repo after the migration.


vxfield (approved):
Looks good to me

PR #40, opened 2021-04-03
Upgrade pygments to 2.7.4

guenp:
Addresses CVE-2021-27291 and CVE-2021-20270

### DIFF
--- a/qdk/setup.py
+++ b/qdk/setup.py
@@ -38,6 +38,7 @@ setuptools.setup(
         'varname',
         'ruamel-yaml',
         'basis_set_exchange',
-        'jupyter_nbextensions_configurator'
+        'jupyter_nbextensions_configurator',
+        'pygments==2.7.4'
     ]
 )
--- a/qdk/setup.py
+++ b/qdk/setup.py
@@ -39,6 +39,6 @@ setuptools.setup(
         'ruamel-yaml',
         'basis_set_exchange',
         'jupyter_nbextensions_configurator',
-        'pygments==2.7.4'
+        'pygments>=2.7.4'
     ]
 )


vxfield (approved):
Shouldn't we try to always use the latest version of these packages instead of pinning to a specific version?
This way if there are new security fixes we the packages we will automatically pick them. There is a risk of breaking changes, but then we can address them as they come, right?

jenshnielsen:
@guenp and @vxfield  its usually a bad idea to pin to an exact version (with ==) in setup.py/setup.cfg. It makes it much harder to deploy the package together with other packages having a requirement on the same package. I would suggest that you do `pygments>=2.7.4` 

In QCoDeS we are using dependabot to keep the requirements that we perform most of the actual ci testing against (in requirements.txt) up to date and automatically open new pull requests when a new package is released. Meanwhile the requirements in setup.py/setup.cfg are normally the minimum supported versions of the required packages. In qcodes we are using requirements-builder to create an additional ci job that test against these. By being flexible on your enforced requirements you make it easier to deploy your package together with other packages.

guenp:
Thanks @jenshnielsen, those are great ideas. I've created a new issue for setting up dependabot: https://github.com/microsoft/qdk-python/issues/41. Jens' suggestion should fix the build error as well.

vxfield:
Now the build issue is unrelated to this PR but an issue with building the qsharp-runtime on Mac...

anjbur (approved):
Confirmed that this resolves the Component Governance flags, looks good to me!

guenp:
Restarted pipeline after https://github.com/microsoft/qsharp-runtime/pull/623 was merged

PR #42, opened 2021-04-14
Fix unit tests imports

vxfield:
This PR fixes imports in the unit tests that were not working in some environments.

After this PR you should be able to run `pytest` and not have any module import issues.

### DIFF
--- a/azure-quantum/tests/unit/test_job.py
+++ b/azure-quantum/tests/unit/test_job.py
@@ -19,7 +19,7 @@ from azure.quantum import Workspace
 from azure.quantum.optimization import Problem
 from azure.quantum.optimization.solvers import SimulatedAnnealing
 from azure.quantum import Job
-from tests.quantum_test_base import QuantumTestBase 
+from quantum_test_base import QuantumTestBase 
 
 class TestJob(QuantumTestBase):
     """TestJob
--- a/azure-quantum/tests/unit/test_optimization.py
+++ b/azure-quantum/tests/unit/test_optimization.py
@@ -15,7 +15,7 @@ import json
 from azure.quantum import Workspace
 from azure.quantum.optimization import Problem, ProblemType, Term
 from azure.quantum.optimization.solvers import ParallelTempering, SimulatedAnnealing, HardwarePlatform, QuantumMonteCarlo
-from tests.quantum_test_base import QuantumTestBase 
+from quantum_test_base import QuantumTestBase 
 
 class TestProblem(QuantumTestBase):
     def test_add_terms(self):
--- a/azure-quantum/tests/unit/test_workspace.py
+++ b/azure-quantum/tests/unit/test_workspace.py
@@ -13,7 +13,7 @@ import json
 import pytest
 
 from azure.quantum import Workspace
-from tests.quantum_test_base import QuantumTestBase 
+from quantum_test_base import QuantumTestBase 
 
 class TestWorkspace(QuantumTestBase):
     def test_workspace_login(self):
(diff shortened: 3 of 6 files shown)


adelebai:
Looks great, thanks for adding these & the pipelines. 
Optional ask: can we update the readme.md with instructions on how to run the unit tests? (i.e. which directory to call from and such).

adelebai (approved):

vxfield:
Thanks, @adelebai. 
I added more information on how to setup the dev environment and run the unit tests in the readmes.

azure-quantum/tests/unit/test_job.py
```diff
@@ -19,7 +19,7 @@
 from azure.quantum.optimization import Problem
 from azure.quantum.optimization.solvers import SimulatedAnnealing
 from azure.quantum import Job
-from tests.quantum_test_base import QuantumTestBase 
+from quantum_test_base import QuantumTestBase
```
sanjgupt, line 22: I think this should be ..quantum_test_base, since the quantum_test_base is one level above unit folder. I get this import failure.

azure-quantum/tests/unit/test_optimization.py
```diff
@@ -15,7 +15,7 @@
 from azure.quantum import Workspace
 from azure.quantum.optimization import Problem, ProblemType, Term
 from azure.quantum.optimization.solvers import ParallelTempering, SimulatedAnnealing, HardwarePlatform, QuantumMonteCarlo
-from tests.quantum_test_base import QuantumTestBase
```
sanjgupt, line 18: should be ..quantum_test_base

azure-quantum/tests/unit/test_workspace.py
```diff
@@ -13,7 +13,7 @@
 import pytest
 
 from azure.quantum import Workspace
-from tests.quantum_test_base import QuantumTestBase
```
sanjgupt, line 16: same as above

azure-quantum/tests/unit/test_job.py
sanjgupt: Actually nvm, i see that you moved the quantum_base_file.py to unit directory. I think that is unnecessary. 
Here's my 2 c: 
1. I think this should be named common.py instead of quantum_test_base.py, just as a clarification of intent of this file. 
2. It should be outside of the unit folder. It may be used for other tests as well. Using  from ..quantum_test_base import QuantumTestBase should get around this.

sanjgupt: All this applies to the remaning comments as well.

vxfield: I would to get rid of the other test folder and only have unit tests that can be run manually, automated, live or in playback mode.
That's why I moved the file here as an initial step.
Makes sense?

I agree with renaming it to common.

sanjgupt: So we will eventually have a flat structure inside /tests where all the uts will be located?

sanjgupt: I am not opposed to this, I am just trying to plan for the changes I am adding now.

vxfield: Yes, we shouldn't need to distinguish the types of tests in different folders.
So I'll go ahead and merge it, but I'm open to suggestion that will come from your PR.

sanjgupt (approved):

sanjgupt:
Approving this and will consume this to restructure the project a little bit and send that PR out in a little bit

PR #49, opened 2021-04-21
Add PA and SSMC solvers

haohaiyu:
Add Population annealing (PA) and substochastic monte carlo (SSMC) solvers to QDK.

new UT tests are added.

This change shall be released in May's schedule.

### DIFF
--- a/azure-quantum/azure/quantum/optimization/solvers.py
+++ b/azure-quantum/azure/quantum/optimization/solvers.py
@@ -28,21 +28,21 @@ __all__ = [
 ]
 
 class RangeSchedule: 
-    def __init__(self, type: str, initial: float, final: float): 
+    def __init__(self, schedule_type: str, initial: float, final: float): 
         """The constructor of Range Scheduler for solver.
 
-        :param type: 
-            specifies type of range scheduler, currently only support 'linear' and 'geometric'.
+        :param schedule_type: 
+            specifies schedule_type of range scheduler, currently only support 'linear' and 'geometric'.
         :param initial: 
             initial value of range schedule.
         :param final:
             stop value of range schedule
         """
-        self.type = type
+        self.schedule_type = schedule_type
         self.initial = initial
         self.final = final
-        if not (self.type == 'linear' or self.type == 'geometric'):
-            raise ValueError('"type" can only be "linear" or "geometric"!')
+        if not (self.schedule_type == 'linear' or self.schedule_type == 'geometric'):
+            raise ValueError('"schedule_type" can only be "linear" or "geometric"!')
 
 class Solver:
 
@@ -147,7 +147,7 @@ class Solver:
         if not(schedule is None):
             if not isinstance(schedule, RangeSchedule):
                 raise ValueError(f'{schedule_name} can only be from class "RangeSchedule"!')
-            schedule_param = {"type": schedule.type, "initial": schedule.initial, "final": schedule.final}
+            schedule_param = {"type": schedule.schedule_type, "initial": schedule.initial, "final": schedule.final}
             self.set_one_param(schedule_name, schedule_param)
 
     def check_set_positive_int(self, var: int, var_name: str):
--- a/azure-quantum/azure/quantum/optimization/solvers.py
+++ b/azure-quantum/azure/quantum/optimization/solvers.py
@@ -16,13 +16,33 @@ from azure.quantum.storage import ContainerClient, create_container_using_client
 logger = logging.getLogger(__name__)
 
 __all__ = [
+    'RangeSchedule',
     'Solver',
     'ParallelTempering',
     'SimulatedAnnealing',
     'HardwarePlatform',
     'Tabu',
-    'QuantumMonteCarlo'
+    'QuantumMonteCarlo',
+    'PopulationAnnealing',
+    'SubstochasticMonteCarlo'
 ]
+
+class RangeSchedule:
+    def __init__(self, type:str, initial:float, final:float):
+        """The constructor of Range Schedulor for solver.
+
+        :param type: 
+            specifies type of range scheduler, currently only support 'linear' and 'geometric'.
+        :param initial: 
+            initial value of range schedule.
+        :param final:
+            stop value of range schedule
+        """
+        self.type = type
+        self.initial = initial
+        self.final = final
+        if not (self.type == 'linear' or self.type == 'geometric'):
+            raise ValueError('"type" can only be "linear" or "geometric"!')
 
 class Solver:
 
@@ -116,6 +136,29 @@ class Solver:
         if value is not None:
             params = self.params["params"] if self.nested_params else self.params
             params[name] = str(value) if self.force_str_params else value
+
+    def check_set_schedule(self, schedule:RangeSchedule, schedule_name:str):
+        if not(schedule is None):
+            if not isinstance(schedule, RangeSchedule):
+                raise ValueError(f'{schedule_name} can only be from class "RangeSchedule"!')
+            schedule_param = {"type":schedule.type, "initial":schedule.initial, "final":schedule.final}
+            self.set_one_param(schedule_name, schedule_param)
+
+    def check_set_positive_int(self, var:int, var_name:str):
+        if not(var is None):
+            if not isinstance(var, int):
+                raise ValueError(f'{var_name} shall be int!')
+            if var <= 0:
+                raise ValueError(f'{var_name} must be positive!')
+            self.set_one_param(var_name, var)
+
+    def check_set_float_limit(self, var:float, var_name:str, var_limit:float):
+        if not(var is None):
+            if not (isinstance(var, float) or isinstance(var, int)):
+                raise ValueError(f'{var_name} shall be float!')
+            if var <= var_limit:
+                raise ValueError(f'{var_name} can not be smaller than {var_limit}!')
+            self.set_one_param(var_name, var)
 
 class HardwarePlatform(Enum):
     CPU = 1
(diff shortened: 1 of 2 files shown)


azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -16,14 +16,34 @@
...
     'HardwarePlatform',
     'Tabu',
-    'QuantumMonteCarlo'
+    'QuantumMonteCarlo',
+    'PopulationAnnealing',
+    'SubstochasticMonteCarlo'
 ]
 
+class RangeSchedule:
+    def __init__(self, type:str, initial:float, final:float):
```
guenp, line 31: Would it be possible to use a different variable name here instead of "type", since it's a protected variable name? something like `range_type` perhaps.
Also, PEP8 guidelines suggest to add spaces after the semicolon in function signatures: https://www.python.org/dev/peps/pep-0008/#variable-annotations

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -16,14 +16,34 @@
...
+        """The constructor of Range Schedulor for solver.
```
guenp, line 32:
```suggestion
        """The constructor of Range Scheduler for solver.
```

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -117,6 +137,29 @@ def set_one_param(self, name: str, value: Any):
             params = self.params["params"] if self.nested_params else self.params
             params[name] = str(value) if self.force_str_params else value
 
+    def check_set_schedule(self, schedule:RangeSchedule, schedule_name:str):
```
guenp, line 140: please add a whitespace after semicolon in accordance with PEP8 guidelines: https://www.python.org/dev/peps/pep-0008/#variable-annotations
This method is missing a docstring.

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -16,14 +16,34 @@
...
     'SimulatedAnnealing',
     'HardwarePlatform',
     'Tabu',
-    'QuantumMonteCarlo'
+    'QuantumMonteCarlo',
+    'PopulationAnnealing',
+    'SubstochasticMonteCarlo'
 ]
 
+class RangeSchedule:
```
guenp, line 30: Some  of the public methods in this class are missing docstrings.

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -117,6 +137,29 @@ def set_one_param(self, name: str, value: Any):
...
             params[name] = str(value) if self.force_str_params else value
 
+    def check_set_schedule(self, schedule:RangeSchedule, schedule_name:str):
+        if not(schedule is None):
+            if not isinstance(schedule, RangeSchedule):
+                raise ValueError(f'{schedule_name} can only be from class "RangeSchedule"!')
+            schedule_param = {"type":schedule.type, "initial":schedule.initial, "final":schedule.final}
+            self.set_one_param(schedule_name, schedule_param)
+
+    def check_set_positive_int(self, var:int, var_name:str):
```
guenp, line 147: same comment here on whitespace + missing docstring.

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -117,6 +137,29 @@ def set_one_param(self, name: str, value: Any):
...
+        if not(var is None):
+            if not isinstance(var, int):
+                raise ValueError(f'{var_name} shall be int!')
+            if var <= 0:
+                raise ValueError(f'{var_name} must be positive!')
+            self.set_one_param(var_name, var)
+
+    def check_set_float_limit(self, var:float, var_name:str, var_limit:float):
```
guenp, line 155: same comment here on whitespace + missing docstring .

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -349,3 +392,97 @@ def __init__(
...
+class PopulationAnnealing(Solver):
+    def __init__(
+        self, 
+        workspace: Workspace, 
+        alpha: Optional[float]=None, 
+        seed: Optional[int]=None,  
+        population: Optional[int] = None,
+        sweeps: Optional[int] = None,
+        beta: Optional[RangeSchedule] = None,
+        culling_fraction: Optional[float]=None
```
guenp, line 404: PEP8 recommends not adding spaces around equal signs: https://www.python.org/dev/peps/pep-0008/#other-recommendations

```suggestion
        alpha: Optional[float]=None, 
        seed: Optional[int]=None,  
        population: Optional[int]=None,
        sweeps: Optional[int]=None,
        beta: Optional[RangeSchedule]=None,
        culling_fraction: Optional[float]=None
```

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -349,3 +392,97 @@ def __init__(
...
+class SubstochasticMonteCarlo(Solver):
+    def __init__(
+        self, 
+        workspace: Workspace, 
+        alpha: Optional[RangeSchedule]=None, 
+        seed: Optional[int]=None,  
+        target_population: Optional[int] = None,
+        step_limit: Optional[int] = None,
+        beta: Optional[RangeSchedule] = None,
+        steps_per_walker: Optional[int]=None
```
guenp, line 452: Please remove spaces around equal signs according to PEP8 guidelines:
https://www.python.org/dev/peps/pep-0008/#other-recommendations

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -117,6 +137,29 @@ def set_one_param(self, name: str, value: Any):
             params = self.params["params"] if self.nested_params else self.params
             params[name] = str(value) if self.force_str_params else value
 
+    def check_set_schedule(self, schedule:RangeSchedule, schedule_name:str):
+        if not(schedule is None):
+            if not isinstance(schedule, RangeSchedule):
+                raise ValueError(f'{schedule_name} can only be from class "RangeSchedule"!')
+            schedule_param = {"type":schedule.type, "initial":schedule.initial, "final":schedule.final}
```
guenp, line 144: Please add spaces after the colons in the dict, see PEP8 guidelines:
https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

```suggestion
            schedule_param = {"type": schedule.type, "initial": schedule.initial, "final": schedule.final}
```

guenp (changes requested):

cgranade:
As a side note, given the number of PEP8 / linting comments, would it perhaps be reasonable to add a pylint pass to CI, similar to the use of formatting and linting tools in the qsharp-compiler and qsharp-runtime CI builds?

azure-quantum/azure/quantum/optimization/solvers.py
haohaiyu: "type" is by design which means the type of class obj, and this is within the class, so shall not be any confusion.
and I could not find the "add spaces after the semicolon in function signatures" in https://www.python.org/dev/peps/pep-0008/#variable-annotations
and the existing code does not have this style too. since there is a newline after semicolon of function, it make no sense to add a space

azure-quantum/azure/quantum/optimization/solvers.py
haohaiyu: fixed, thanks!

azure-quantum/azure/quantum/optimization/solvers.py
haohaiyu: docstring are added.

guenp:
Thanks @cgranade for the suggestion! @vxfield is planning to do that this week, issue to follow.

guenp:
@haohaiyu @vxfield I've created an issue to add PEP8 and type checking to the pipeline: https://github.com/microsoft/qdk-python/issues/50

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -16,14 +16,34 @@
...
     'HardwarePlatform',
     'Tabu',
-    'QuantumMonteCarlo'
+    'QuantumMonteCarlo',
+    'PopulationAnnealing',
+    'SubstochasticMonteCarlo'
 ]
 
+class RangeSchedule: 
+    def __init__(self, type: str, initial: float, final: float):
```
guenp, line 31:
```suggestion
    def __init__(self, range_type: str, initial: float, final: float): 
```

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -16,14 +16,34 @@
...
+        """The constructor of Range Scheduler for solver.
+
+        :param type: 
+            specifies type of range scheduler, currently only support 'linear' and 'geometric'.
+        :param initial: 
+            initial value of range schedule.
+        :param final:
+            stop value of range schedule
+        """
+        self.type = type
```
guenp, line 41:
```suggestion
        self.type = range_type
```

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -16,14 +16,34 @@
...
+    'QuantumMonteCarlo',
+    'PopulationAnnealing',
+    'SubstochasticMonteCarlo'
 ]
 
+class RangeSchedule: 
+    def __init__(self, type: str, initial: float, final: float): 
+        """The constructor of Range Scheduler for solver.
+
+        :param type:
```
guenp, line 34:
```suggestion
        :param range_type: 
```

guenp (changes requested):

azure-quantum/azure/quantum/optimization/solvers.py
haohaiyu: the type is for the whole class obj, range_type will be confusing.

azure-quantum/azure/quantum/optimization/solvers.py
haohaiyu: explained. will not fix.

azure-quantum/azure/quantum/optimization/solvers.py
vxfield: what about `scheduler_type` or something else less generic than just `type`?

vxfield (approved):
Approved with suggestion, as Guen pointed out, that it might be better to use a less generic parameter name then `type` in the `RangeScheedule` constructor.

guenp:
@vxfield Thanks for reviewing. I highly recommend not using variable names that overlap with built-in variable names.
See also this stackoverflow post: https://stackoverflow.com/questions/42378996/variable-names-to-avoid-in-python

azure-quantum/azure/quantum/optimization/solvers.py
guenp: I've seen people use `_type` also.

vxfield (commented):
I'm retracting my approval since Guen has way more experience than I with Python and she thinks we should really avoid using the parameter named `type`.

vxfield (changes requested):
Haohai, please follow Guen's recommendation of not using the parameter named `type`

guenp (approved):
Thanks for making the changes!

PR #53, opened 2021-04-29
Fix PEP8 code style and misspelling issues

vxfield:
This PR fixes hundreds of PEP8/Pylint and some misspelling issues in preparation #50 (enforce code style guidelines in the CI/CD pipeline)

PR #55, opened 2021-05-03
Sanjgupt/onlineproblem

sanjgupt:
Create online problem for problem reuse and add client side warning etc. 
The design diagram can be found here: 
https://dev.azure.com/ms-quantum/Quantum%20Program/_wiki/wikis/Quantum-OKRs.wiki/2009/Problem-reuse-behavior


azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -14,8 +14,8 @@
 from typing import List, Union, Dict, Optional, TYPE_CHECKING
 from enum import Enum
 from azure.quantum.optimization import Term
-from azure.quantum.storage import upload_blob, ContainerClient
-
+from azure.quantum.storage import upload_blob, ContainerClient, BlobClient, download_blob
+#import azure.quantum.storage
```
adelebai, line 18: nit- is this comment still needed?

azure-quantum/azure/quantum/optimization/online_problem.py
```diff
@@ -0,0 +1,27 @@
...
+if TYPE_CHECKING:
+    from azure.quantum.workspace import Workspace
+
+class OnlineProblem(Problem):
+    def __init__(self, name:str, blob_uri:str):
+        super().__init__(self)
+        self.name = name
+        self.uploaded_blob_uri = blob_uri
+    
+    def evaluate(self, configuration: Union[Dict[int, int], Dict[str, int]]) -> float:
```
adelebai, line 18: Might need a similar error message for set_fixed_variables too I think.

adelebai: Also in line with the above comment, if the problem isn't downloaded, do we need to check & throw exceptions for all the existing functionality that rely on a downloaded problem?

azure-quantum/azure/quantum/optimization/online_problem.py
```diff
@@ -0,0 +1,33 @@
...
+from azure.quantum.optimization import Problem
+
+logger = logging.getLogger(__name__)
+
+__all__ = ['OnlineProblem']
+
+if TYPE_CHECKING:
+    from azure.quantum.workspace import Workspace
+
+class OnlineProblem(Problem):
```
adelebai, line 12: As mentioned offline, we could consider the possibility of doing this the other way around - i.e. having Problem inherit OnlineProblem instead. This way we don't need to throw all those exceptions for unsupported methods.

adelebai: On second thought, maybe they should both inherit from a common base (?) and then OnlineProblem can still call Problem  - I am not sure, let me know your thoughts.

vxfield: +1 on having a ProblemBase that is an abstract class and have the Problem and OnlineProblem being descends of it.

adelebai: If the .download() functionality is giving you too much trouble, I'm happy to skip it and just let Problem.py take in a blob url or something if the user wants to explicitly download a blob into a Problem.py. In terms of just repeat submissions, I'm happy with OnlineProblem being a wrapper around a blob uri & name

vxfield: I suggest waiting for #53 before doing more changes because #53 has a lot of PEP8 fixes across many files.

sanjgupt: In order to resolve this, I have made OnlineProblem not inherit from problem and use Problem's download method to download it. Please let me know if this looks better now. 
We can visit the idea of having a base problem class and have all three types of problems (Problem, OnlineProbem and Streaing Problem inherit from it) but that might tie into larger changes more refactiring.

azure-quantum/azure/quantum/optimization/online_problem.py
```diff
@@ -0,0 +1,22 @@
...
+        blob_uri: str,
+        **kw
+        ):
+        super(OnlineProblem, self).__init__(**kw)
+        self.name = name
+        self.uploaded_blob_uri = blob_uri
+
+    def download(self) -> Problem:
+        logger.warning("The problem will be downloaded to the client")
+        return Problem.download(self)
```
adelebai, line 22: I just realized - would this actually work with the "self" parameter? Since these are 2 different classes, and Problem.download technically doesn't take in a parameter? 

I think download_blob(self.uploaded_blob_uri) and Problem.deserialize(contents, self.name) might need to be called at this level

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -275,5 +279,30 @@ def is_large(self) -> bool:
...
+        a list of terms with that index
+        """
+        terms = []
+        if self.terms != []:
+            for term in self.terms:
+                if id in term.ids:
+                    terms.append(term)
+            return terms
+        else:
+            raise Exception("Please download the problem on the client" \
```
adelebai, line 307: nit: Adding "There are currently no terms in this problem." at the beginning might give more context. So we have something like

"There are currently no terms in this problem. Please download the problem on the client or add terms to the problem to perform this operation."

azure-quantum/azure/quantum/optimization/online_problem.py
sanjgupt: The way I understood it was self is the object and calling Problem.download as a class method can enable taking the object as parameter. But actually let me test again.

sanjgupt: Hi, I updated the download function to handle authetication correctly. This still works with the onlineproblem when Problem.download is called as a class method with the online problem object passed to self.

adelebai (approved):

adelebai:
Looks good to me. I'm surprised the download() requires workspace auth - I was assuming that customers would be calling it with the full uri (including access token), but in case they aren't I think the current approach works too.

vxfield:
Hi @sanjgupt, can you revert your changes to the `_client` folder?
Those files should not be changed manually (even if it's for code styling issues) as they are auto-generated and can be replaced at any time.
#54 will soon update the `_client` folder anyway.

azure-quantum/tests/unit/test_online_problem.py
```diff
@@ -0,0 +1,38 @@
...
+import json
+
+
+class TestOnlineProblemClass(unittest.TestCase):
+    def setUp(self):
+        self.mock_ws = Mock()
+        self.mock_ws._get_linked_storage_sas_uri.return_value = Mock()
+        self.o_problem = OnlineProblem(name="test", blob_uri="mock_blob_uri")
+
+    @pytesat.skip(reason= "OnlineProblem not available in SDK yet")
```
guenp, line 25:
```suggestion
    @pytest.skip(reason= "OnlineProblem not available in SDK yet")
```

PR #61, opened 2021-05-13
Add translation from NPZ to Terms

frtibble:
Adds a utility function `terms_from_npz()` to `problem.py`:
- includes helper function `is_valid_npz` to check whether the supplied NPZ file's array names match the default or user-supplied namings
- adds `terms_from_npz` which loads an NPZ file, reads in the data, and converts it to a list of `Term` (which can be added to a problem)
- adds unit tests to `test_problem.py`, this tests the loading of qubo and pubo problems from an NPZ file and their translation into Terms

### DIFF
--- a/azure-quantum/azure/quantum/optimization/problem.py
+++ b/azure-quantum/azure/quantum/optimization/problem.py
@@ -9,6 +9,8 @@ import uuid
 import io
 import gzip
 import json
+import numpy
+import os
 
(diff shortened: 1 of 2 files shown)


frtibble:
Hi @adelebai @vxfield - requesting your review when it's convenient, thank you!

vxfield (approved):
Looks good to me.

PR #64, opened 2021-05-19
Serialize problem name

vxfield:
This PR serializes the problem name in the json payload and deserializes it.

The deserialization method now has the `name` parameter as optional, since it will always try to deserialize the problem name from the json payload.
If `name` is specified during the `.deserialize` method, it will be used ignoring the problem name from the json.

The `name` is stored in a new **optional** metadata field at root of the problem schema.


azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -74,6 +74,7 @@ def __init__(
     def serialize(self) -> str:
         """Serializes the problem to a JSON string"""
         result = {
+            "name": self.name,
```
sanjgupt, line 77: If we return the name like this, then this file is not directly usable in qio. i.e. a user cannot submit this file to run qio problems because it changes the input file format schema. Is this the intended affect?

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -87,17 +88,26 @@ def serialize(self) -> str:
...
-        :type name: str
+        :param name: 
+            The name of the problem is optional, since it will try 
+            to read the serialized name from the json payload.
+            If this parameter is not empty, it will use it as the
+            problem name ignoring the serialized value.
+        :type name: Optional[str]
         """
         result = json.loads(problem_as_json)
+        name = name if name else result.get("name")
```
sanjgupt, line 110: In line with the above comment, a name for the problem is not a part of the problem schema.

azure-quantum/azure/quantum/optimization/problem.py
vxfield: I thought it would be backwards compatible because it's just a new json parameter that is optional and new json parameters should cause problems, but I think you are right that this may affect consumers that have a more rigid parser.
Do we have another place to store the name that is safe?

vxfield: We decided to introduce a new **optional** `metadata` field to store the name, so resolving this comment.

sanjgupt (approved):

azure-quantum/tests/unit/test_job.py
```diff
@@ -166,12 +166,16 @@ def test_job_upload_and_run_solvers(self):
...
 
-                # Check if job succeeded during live tests
-                if not self.is_playback:
-                    job.refresh()
-                    job.get_results()
-                    assert job.has_completed()
-                    assert job.details.status == "Succeeded"
+                # For recording purposes, we only want to record and
+                # and resume recording when the job has completed
+                self.pause_recording()
```
guenp, line 171: This is a very useful feature, thanks for adding!

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -104,18 +107,31 @@ def serialize(self) -> str:
...
+            If this parameter is not empty, it will use it as the
+            problem name ignoring the serialized value.
+        :type name: Optional[str]
         """
         result = json.loads(problem_as_json)
-
+        
+        if (name == None):
+            metadata = result.get("metadata")
+            if (metadata != None):
```
guenp, line 132:
```suggestion
        if name == None:
            metadata = result.get("metadata")
            if metadata != None:
```

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -104,18 +107,31 @@ def serialize(self) -> str:
         return json.dumps(result)
 
     @classmethod
-    def deserialize(cls, problem_as_json: str, name: str):
+    def deserialize(
+            cls, 
+            problem_as_json: str, 
+            name: Optional[str] = None
```
guenp, line 113: This is a breaking change, if I don't pass a `name` here and if the `problem_as_json` doesn't have a name, then the problem will not have a name; previously this scenario would fail because `name` is a required parameter. Should we throw a `ValueError` if `name` is `None` both as input argument and in the problem JSON?

guenp (approved):
I left a few comments, but otherwise LGTM. It'd be great to add some documentation on the new `pause_recording` feature. Thanks!

azure-quantum/azure/quantum/optimization/problem.py
vxfield: Thanks for catching it.
I have handled the None value in the Problem constructor, defaulting it to "Optimization problem".

anpaz (approved):
lgtm

PR #71, opened 2021-06-11
Update docstrings for new solvers.

andrist:
Docstring for SSMC was copy-pasted but not updated fully (name was still PA).

This PR:
- Fixes the copy-pasted name for SSMC
- Reorders parameters similar to other solvers (starting with seed, sweeps)
- Harmonizes capitalization in the param description (some were capitalized some not)
- Fixes some missing/superfluous articles ("the") and ("a" -> "an")
- Provides more details for the parameters of type RangeSchedule
- Removes the culling_fraction parameter (it doesn't do anything when using this API).

NOTE: "target_population" should be changed to "population" for SSMC, but this will be addressed in a separate PR.


guenp:
@andrist would you mind resolving the branch conflicts and retriggering a build?

anpaz:
These changes have been incorporated into the repo already.

PR #72, opened 2021-06-11
Make the credential a required parameter to create a workspace

vxfield:
This PR makes the credential a required parameter to create a workspace, per Azure SDK recommendations.

Users will need to specify a valid Azure Credential from https://docs.microsoft.com/en-us/python/api/overview/azure/identity-readme?view=azure-python#credential-classes.

There will be another PR soon to create a new page in our docs with examples and the https://aka.ms/quantum-python-sdk-auth link will point to it.


vxfield:
Closing this PR in favor of #69 which will offer better customer experience for authentication.

PR #79, opened 2021-06-25
Add IonQ provider with JSON passthrough

guenp:
Todo's:
[x] add unit tests
[x] refactor `Job` (see PR https://github.com/microsoft/qdk-python/pull/84)
[x] fix UT recording issue


azure-quantum/azure/quantum/providers/ionq.py
```diff
@@ -0,0 +1,75 @@
...
+IONQ_PROVIDER_ID = "IonQ"
+DEFAULT_TIMEOUT = 100
+
+
+class IonQ:
+    def __init__(self, workspace: Workspace, target: str = "ionq.simulator"):
+        self.workspace = workspace
+        self.target = target
+    
+    def submit(
```
anpaz, line 22: I was hoping to push most of this functionality to the Job class itself. For example uploading the Blob and populating the JobDetails.
Then we can clean the ../optimization/Solver code to use the same logic, so we don't have that logic duplicated.

azure-quantum/azure/quantum/providers/ionq.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+        circuit: IonQJson,
```
anpaz, line 24: The circuit parameter should be some opaque datastructure (byte[], stream) or some generic interface that just implements a `serialize` method such that it can be used regardless of how the actual format the provider expects. 
Even more important if we move this method to the `Job` class as I mentioned in my other comment. As a Job can have any type of payload.

azure-quantum/azure/quantum/providers/ionq.py
```diff
@@ -0,0 +1,75 @@
...
+        job = self.workspace.submit_job(job)
+        return job
+
+    def encode_data(circuit: IonQJson) -> bytes:
+        data = io.BytesIO()
+        circuit = json.dumps(circuit)
+        data.write(circuit.encode())
+        return data.getvalue()
+
+    def upload_blob(
```
anpaz, line 63: this method should be moved to Job too.

azure-quantum/azure/quantum/providers/ionq.py
```diff
@@ -0,0 +1,75 @@
...
+            input_data_uri=uploaded_blob_uri,
+            provider_id=IONQ_PROVIDER_ID,
+            target=self.target,
+            input_params={'params': {'timeout': timeout}},
+        )
+        job = Job(self.workspace, details)
+        job = self.workspace.submit_job(job)
+        return job
+
+    def encode_data(circuit: IonQJson) -> bytes:
```
anpaz, line 57: This should really be in the Payload itself, imho.

anpaz (changes requested):
We need to add tests that can run both recorded and live here:
https://github.com/microsoft/qdk-python/tree/guenp/feature/ionq-pass-through/azure-quantum/tests/unit

We also need to refactor some of the optimization classes to remove code duplication.

If you rather do these in a different PR let me know.

azure-quantum/azure/quantum/providers/ionq.py
guenp: Do you think `Payload` should be a separate class?

guenp:
@anpaz, I've done the `Job` refactor in this PR: https://github.com/microsoft/qdk-python/pull/84

azure-quantum/tests/unit/test_job.py
```diff
@@ -96,48 +96,89 @@ def get_test_job_id(self):
...
+        workspace = self.create_workspace()
+
+        with unittest.mock.patch.object(
+            Job,
+            self.mock_create_job_id_name,
+            return_value=self.get_test_job_id(),
+        ):
+            workspace = self.create_workspace()
+            blob = problem.to_blob(compress=False)
+            job = Job.from_input_data(
```
anpaz, line 140: I wouldn't use `Job.from_input_data` as the mechanism to upload input_data if  you want to use it on multiple jobs.  I would use the `problem.upload` method to upload the Problem, then I would move creating the job instead to the `for` loop below as you really want to create one Job per solver.

azure-quantum/azure/quantum/target/target.py
```diff
@@ -0,0 +1,90 @@
...
+        """Implement abstract method to encode input data to bytes
+
+        :param data: Input data
+        :type data: Dict[Any, Any]
+        :return: Encoded input data
+        :rtype: bytes
+        """
+        pass
+
+    def submit_input_data(
```
anpaz, line 43: I don't see the value of this method. Why don't we just have `submit`?

guenp: This method takes a `bytes` object while `submit` takes the deserialized version, e.g. JSON `dict`

azure-quantum/tests/unit/test_job.py
guenp: Good call, thanks!

azure-quantum/azure/quantum/target/target.py
guenp: I'll make this a private method for now.

guenp:
UT recording issues are fixed in https://github.com/microsoft/qdk-python/pull/98

anpaz (approved):
lgtm!

Issue #80, opened 2021-07-02
Authentication issue with MSAL Credentials on Windows: ImportError: DLL load failed while importing win32file: The specified module could not be found.


vxfield:
### Summary
On Windows, when trying to import the `azure.quantum` module (for example: `from azure.quantum import Workspace`) it fails with `ImportError: DLL load failed while importing win32file: The specified module could not be found.`.

### Details
The import fails when trying to use `SharedTokenCacheCredential` and internally importing the `MSAL-Extensions` and `PortaLocker` packages with a stack-trace ending like the following:

```text
File "C:\Users\username\Miniconda3\envs\qiot\lib\site-packages\portalocker\portalocker.py", line 9, in <module>
    import win32file
ImportError: DLL load failed while importing win32file: The specified module could not be found.
````

### Cause
You likely have an older version of `azure-identity` and `pywin32` packages that has a [bug](https://github.com/Azure/azure-sdk-for-python/issues/8292).

### Solution
In the July 2021 release we will update the dependencies of the `azure-quantum` package to require minimum versions of our dependencies.

### Workaround
Make sure to install the latest `azure-identity` package in your environment.
```bash
pip install azure-identity --upgrade
```

vxfield:
Fixed with #81

Issue #41, opened 2021-04-12
Set up dependabot to automatically keep dependency versions up to date


guenp:
See comment by @jenshnielsen here: https://github.com/microsoft/qdk-python/pull/40#issuecomment-816473328

We should set up [dependabot](https://dependabot.com/) to keep the requirements that we perform most of the actual ci testing against (in requirements.txt) up to date and automatically open new pull requests when a new package is released. See also the [Qcodes](https://github.com/QCoDeS/Qcodes) repo for how this is set up.

hxri:
Hi Is this done? If not can I work on it?

guenp:
Hi @hxri! No not yet, if you're still interested please do! I'll add a Hacktoberfest label as well.

PR #81, opened 2021-07-02
azure-quantum: Require minimum version of dependencies

vxfield:
This PR add requirements for minimum versions of the dependencies needed by the `azure-quantum` package.

We should eventually use Dependabot for this #41, but at least this is a start and if fixes issue #80.


guenp (approved):
Thanks @vxfield for adding these!

PR #82, opened 2021-07-06
Add limit checks for schedule

andrist:
Add checks for the evolution (INCREASING vs DECREASING) to the schedule parameter setter.

This allows misconfigurations of the schedules for alpha, beta in PopulationAnnealing and
SubstochasticMonteCarlo to be registered (and thrown) client side. These are already
checked server-side.

(From email thread, no ADO item).

Plus various cleanups of docstrings and requested refactoring of variable names.

### DIFF
--- a/azure-quantum/azure/quantum/optimization/solvers.py
+++ b/azure-quantum/azure/quantum/optimization/solvers.py
@@ -585,6 +585,7 @@ class SubstochasticMonteCarlo(Solver):
         step_limit: Optional[int] = None,
         beta: Optional[RangeSchedule] = None,
         steps_per_walker: Optional[int] = None,
+        timeout: Optional[int] = None,
     ):
         """The constructor of Substochastic Monte Carlo solver.
 
@@ -606,8 +607,17 @@ class SubstochasticMonteCarlo(Solver):
             beta (resampling factor) values evolve over time
         :param steps_per_walker:
             number of steps to attempt for each walker, must be postive
+        :param timeout:
+            specifies maximum number of seconds to run the core solver
+            loop. Initialization time does not respect this value, so the
+            solver may run longer than the value specified. Setting this value
+            will trigger the parameter free substochastic monte carlo solver.
         """
-        target = "microsoft.substochasticmontecarlo.cpu"
+
+        if timeout is None:
+            target = "microsoft.substochasticmontecarlo.cpu" 
+        else:
+            target = "microsoft.substochasticmontecarlo-parameterfree.cpu"
         super().__init__(
             workspace=workspace,
             provider="Microsoft",
@@ -621,3 +631,4 @@ class SubstochasticMonteCarlo(Solver):
         self.check_set_positive_int(target_population, "target_population")
         self.check_set_positive_int(steps_per_walker, "steps_per_walker")
         self.check_set_positive_int(step_limit, "step_limit")
+        self.check_set_positive_int(timeout, "timeout")
--- a/azure-quantum/azure/quantum/optimization/solvers.py
+++ b/azure-quantum/azure/quantum/optimization/solvers.py
@@ -234,15 +234,15 @@ class Solver:
         the specified bound.
 
         :param schedule_name:
-            name of the schedule parameter.
+            Name of the schedule parameter.
         :param schedule_value:
-            schedule value to be assigned and checked.
+            Schedule value to be assigned and checked.
         :param evolution
-            expected schedule evolution (INCREASING or DECREASING)
+            Expected schedule evolution (INCREASING or DECREASING)
         :lower_bound_exclusive:
-            exclusive lower bound for both ends of the schedule, optional.
+            Exclusive lower bound for both ends of the schedule, optional.
         :lower_bound_inclusive:
-            inclusive lower bound for both ends of the schedule, optional.
+            Inclusive lower bound for both ends of the schedule, optional.
         """
         if not (schedule_value is None):
             if not isinstance(schedule_value, RangeSchedule):
@@ -294,9 +294,9 @@ class Solver:
         and check that it is a positive integer.
 
         :param parameter_name:
-            name of the parameter.
+            Name of the parameter.
         :param parameter_value:
-            value to be assigned and checked.
+            Value to be assigned and checked.
         """
         if not (parameter_value is None):
             if not isinstance(parameter_value, int):
@@ -321,13 +321,13 @@ class Solver:
         and check that it has a float value satisfying bounds.
 
         :param parameter_name:
-            name of the parameter.
+            Name of the parameter.
         :param parameter_value:
-            value to be assigned and checked.
+            Value to be assigned and checked.
         :lower_bound_exclusive:
-            exclusive lower bound to check parameter_value against, optional.
+            Exclusive lower bound to check parameter_value against, optional.
         :lower_bound_inclusive:
-            inclusive lower bound to check parameter_value against, optional.
+            Inclusive lower bound to check parameter_value against, optional.
         """
         if not (parameter_value is None):
             if not (isinstance(parameter_value, float) or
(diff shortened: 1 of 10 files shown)


azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -214,12 +214,27 @@ def check_submission_warnings(self, problem: Problem):
                         and resubmitting with a lower timeout."
                 )
 
-    def check_set_schedule(self, schedule: RangeSchedule, schedule_name: str):
+    class ScheduleEvolution(Enum):
+        INCREASING = 1
+        DECREASING = 2
+
+    def check_set_schedule(
+        self, schedule: RangeSchedule, schedule_name: str,
```
guenp, line 222: Please start new variables on a newline (as in the suggestion below) for all other methods as well, thanks!
```suggestion
        self,
        schedule: RangeSchedule,
        schedule_name: str,
```

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -248,25 +288,60 @@ def check_set_positive_int(self, var: int, var_name: str):
             self.set_one_param(var_name, var)
 
     def check_set_float_limit(
-        self, var: float, var_name: str, var_limit: float
+        self, var: float, var_name: str,
+        lower_bound_exclusive: Optional[float] = None,
+        lower_bound_inclusive: Optional[float] = None,
     ):
-        """Check whether the var parameter is a float larger than var_limit.
+        """Check whether the var parameter is a float satisfying bounds.
```
guenp, line 295: What is "var"? Is this "variable" or "variance"? And if it's the former could we come up with a more descriptive name?

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -248,25 +288,60 @@ def check_set_positive_int(self, var: int, var_name: str):
...
+            self.set_one_param(var_name, var)
+
+    def check_limit(
+        self, var: Optional[float], var_name: str,
+        lower_bound_exclusive: Optional[float] = None,
+        lower_bound_inclusive: Optional[float] = None,
+    ):
+        """Check whether the var parameter is a float satisfying bounds.
+        :param var:
+            var paramter to be checked
```
guenp, line 321: Typo here; also what is "the var parameter" (same question as above)?

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -533,29 +608,28 @@ def __init__(
...
-        """The constructor of Population Annealing Search solver.
+        """Constructor of the Population Annealing solver.
 
         Population Annealing Search solver for binary optimization problems
         with k-local interactions on an all-to-all graph topology with double
         precision support for the coupler weights.
 
-        This solver is CPU only, and not support parameter free now.
+        This solver is CPU only.
+        It currently does not support parameter-free invocation.
```
guenp, line 619: Thanks for catching these!

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -533,29 +608,28 @@ def __init__(
...
-            specifies a random seed value.
-        :population:
-            size of target population, must be positive
+            Specifies the random number generator seed value.
         :param sweeps:
-            Number of monte carlo sweeps
+            Number of Monte Carlo sweeps. Must be positive.
+        :population:
+            Size of the population. Must be positive.
+        :param alpha:
```
guenp, line 627: It looks like the order of the arguments in the function signature does not match the order of the arguments in the docstring. For consistency and ease of maintenance I'd recommend to match the order here.

guenp (changes requested):
Some suggested changes, thanks! Some context in the description or a linked issue would be helpful for me to understand the purpose of the PR, if you don't mind.

azure-quantum/azure/quantum/optimization/solvers.py
andrist: not introduced in this PR, but fixed this copy pasted section and both of the original occurrences in line 296ff and 282ff

azure-quantum/azure/quantum/optimization/solvers.py
andrist: This is the original name of the argument; I did not introduce it.
It seems to be used as a synonym for parameter in this code.
I changed all occurrences to `parameter_value` and `parameter_name` instead.

In my opinion, this is not the type of cleanup to request in a feature PR, it belongs in its own PR.

azure-quantum/azure/quantum/optimization/solvers.py
andrist: personally, I find this an unintuitive order for the arguments (and I would prefer there to be more correspondence when it comes to shared arguments with other solvers). I did not change the order however, since users might have code calling these constructors with positional arguments (i.e., would be a client-facing change).

azure-quantum/azure/quantum/optimization/solvers.py
guenp: @andrist thanks for doing that, I agree this would have been fine to do in a separate PR; I wasn't use if you had introduced that argument or not. The effort is still much appreciated!

azure-quantum/azure/quantum/optimization/solvers.py
guenp: Thank you!

azure-quantum/azure/quantum/optimization/solvers.py
guenp: Agreed, though aside from it being a breaking change it still might be worth refactoring this in a separate PR; if you think it would be valuable to do so feel free to file an issue with your preferred argument ordering and we can discuss/look into it.

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -214,59 +217,162 @@ def check_submission_warnings(self, problem: Problem):
...
+        :param evolution
+            Expected schedule evolution (INCREASING or DECREASING)
+        :lower_bound_exclusive:
+            Exclusive lower bound for both ends of the schedule, optional.
+        :lower_bound_inclusive:
+            Inclusive lower bound for both ends of the schedule, optional.
         """
-        if not (schedule is None):
-            if not isinstance(schedule, RangeSchedule):
+        if not (schedule_value is None):
```
guenp, line 247:
```suggestion
        if schedule_value is not None:
```

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -214,59 +217,162 @@ def check_submission_warnings(self, problem: Problem):
...
-        if not (var is None):
-            if not isinstance(var, int):
-                raise ValueError(f"{var_name} shall be int!")
-            if var <= 0:
-                raise ValueError(f"{var_name} must be positive!")
-            self.set_one_param(var_name, var)
-
-    def check_set_float_limit(
-        self, var: float, var_name: str, var_limit: float
+        if not (parameter_value is None):
```
guenp, line 332:
```suggestion
        if parameter_value is not None:
```

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -214,59 +217,162 @@ def check_submission_warnings(self, problem: Problem):
...
+        :lower_bound_exclusive:
+            Exclusive lower bound to check parameter_value against, optional.
+        :lower_bound_inclusive:
+            Inclusive lower bound to check parameter_value against, optional.
         """
-        if not (var is None):
-            if not (isinstance(var, float) or isinstance(var, int)):
-                raise ValueError(f"{var_name} shall be float!")
-            if var <= var_limit:
+        if not (parameter_value is None):
```
guenp, line 361:
```suggestion
        if parameter_value is not None:
```

guenp (approved):
LGTM, thanks for doing the extra cleanup/renaming!

PR #84, opened 2021-07-07
Consolidate blob upload, submit methods

guenp:
Fixes https://github.com/microsoft/qdk-python/issues/83.
Refactored `Job` and added generic blob upload/submit methods.

@anpaz `Solver.submit` can optionally be refactored further if we are OK with deprecating `Problem.upload()`, it would look something like this:
```python
if isinstance (problem, Problem):
    # Upload problem to container storage
    blob = problem.to_blob(compress=compress)
    job = Job.from_blob(
        workspace=self.workspace,
        name=problem.name,
        target=self.target,
        blob=blob,
        blob_name="inputData",
        encoding="gzip" if compress else ""
    )
    job.submit()
```
CC @vxfield @frtibble
we can do that in this PR or a follow up PR later. LMK what you think.


guenp:
@anpaz, as mentioned in https://github.com/microsoft/qdk-python/issues/83 I wanted to propose to move the `submit()` logic to `Workspace`, we can do that in a follow up PR.

guenp:
Updated base to feature branch which we can merge to `main` once we align on design details.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+    from azure.quantum.workspace import Workspace
+
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_TIMEOUT = 100
+
+
+class BaseJob(abc.ABC):
+    # Optionally override these to create a Provider-specific Job subclass
```
anpaz, line 28: What is this `abc.ABC`?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+
+    @property
+    def container_name(self):
+        return f"job-{self.id}"
+
+    @classmethod
+    def from_input_data(
+        cls,
+        workspace: "Workspace",
+        name: str,
```
anpaz, line 50: should this be just `Workspace` (no quotes?)

azure-quantum/azure/quantum/job/base_job.py
guenp: It's an abstract base class, this means you can't instantiate an object from this class (you have to use a subclass). For more info see: https://docs.python.org/3/library/abc.html

azure-quantum/azure/quantum/job/base_job.py
guenp: I made this a string to avoid a circular import.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+            name=name,
+            input_data_format=input_data_format,
+            output_data_format=output_data_format,
+            provider_id=provider_id,
+            input_params=input_params
+        )
+
+    @classmethod
+    def from_uri(
+        cls,
```
anpaz, line 115: it's not clear that `uri` here refers to an existing storage ur, so maybe name this `from_storage_uri`)?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+        name: str,
+        workspace: "Workspace",
+        job_id: str,
+        target: str,
```
anpaz, line 119: Can we make the `job_id` optional, and automatically create one if `None`?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+        :type provider_id: str, optional
+        :param input_params: Input parameters, defaults to None
+        :type input_params: Dict[str, Any], optional
+        :param input_params: Input params for job
+        :type input_params: Dict[str, Any]
+        :return: Azure Quantum Job
+        :rtype: Job
+        """
+        job_id = cls.create_job_id()
+        container_name = f"job-{job_id}"
```
anpaz, line 88: can we make `job_id` an optional parameter to make this consistent with `from_uri`?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+            output_data_format=output_data_format,
+            provider_id=provider_id,
+            input_params=input_params
+        )
+
+    @classmethod
+    def from_uri(
+        cls,
+        name: str,
+        workspace: "Workspace",
```
anpaz, line 117: can we make `workspace` the first parameter? and also, can we make sure the order of the arguments is consistent between the two methods?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+        :param provider_id: Provider ID, defaults to None
+        :type provider_id: str, optional
+        :param input_params: Input parameters, defaults to None
+        :type input_params: Dict[str, Any], optional
+        :return: Job instsance
+        :rtype: Job
+        """
+        if input_params is None:
+            input_params = {"params": {"timeout": DEFAULT_TIMEOUT}}
+        if input_data_format is None:
```
anpaz, line 155: this is actually not universal for all targets, so can't be set as default.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+                workspace.storage, container_name
+            )
+        return container_uri
+
+    @staticmethod
+    def upload_input_data(
+        container_uri: str,
+        input_data: bytes,
+        input_data_name = "inputData",
+        content_type = "application/json",
```
anpaz, line 204: I would call this "blob_name"

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+        encoding = "",
```
anpaz, line 205: I'm not sure this method can assume anything about the content_type, so I wouldn't make this optional.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -0,0 +1,240 @@
...
+        return_sas_token: bool = False
+    ) -> str:
```
anpaz, line 207: It's unclear what this parmater is for or if it's really needed.

anpaz:
>                 self.workspace,

Instead of calling `upload`, I would call the `from_input_data` 

---
Refers to: azure-quantum/azure/quantum/optimization/solvers.py:108 in e883fe4. [](commit_id = e883fe4d62ad1589d972d659a430daef0440ca7f, deletion_comment = False)

anpaz (approved):
i left a couple of comments, but overall lgtm.
I am pretty sure we still need a way to upload a Problem to storage and reuse this URL to submit multiple jobs; so I don't think we can deprecate `upload` itself but maybe we can simplify the code otherwise (see my comment).

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -95,29 +96,8 @@ def submit(
...
-        else:
-            # Storage account is passed, use it to generate a container_uri
-            container_uri = azure.quantum.storage.get_container_uri(
-                self.workspace.storage, container_name
-            )
-
-        logger.debug(f"Container URI: {container_uri}")
+        container_uri = Job.create_container(self.workspace, container_name)
 
         if isinstance(problem, str):
```
anpaz, line 102: Instead of calling `problem.upload`, one thing you could do is call `Job.from_input_data` if the problem is `Problem` and `Job.from_uri`, if using a storage URL, that would remove all the logic of creating the container and job id from this method and pus it to `Job`

azure-quantum/azure/quantum/job/base_job.py
guenp: I'll default to an empty `dict` in that case.

azure-quantum/azure/quantum/job/base_job.py
guenp: This is optional, for removing the sas token from the blob URI or not. I'll add a docstring.

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -92,66 +93,53 @@ def submit(
...
+                provider_id=self.provider,
+                input_data_format=self.input_data_format,
+                output_data_format=self.output_data_format,
+                input_params=self.params,
             )
+        
         else:
-            name = problem.name
-            problem_uri = problem.uploaded_blob_uri
+            if hasattr(problem, "uploaded_blob_uri"):
```
anpaz, line 113: This is a nice enhancement, the docstring for problem should be updated to reflect it, though.

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -92,66 +93,53 @@ def submit(
...
-            problem_uri = problem
-        elif isinstance(problem, Problem):
-            name = problem.name
-            problem_uri = problem.upload(
-                self.workspace,
-                compress=compress,
-                container_name=container_name,
+        if isinstance(problem, Problem):
+            # Create job from input data
+            blob = problem.to_blob(compress=compress)
```
anpaz, line 98: i'm thinking that the `to_blob` should return a tuple or some other datastructure that indicates not only the payload, but also the `content_type`, `encoding` and maybe even the `input_data_type` as all that information is tightly coupled.
On a similar way, maybe instead of accepting just a storage_uri as parameter, the method accepts the same datastructure?

anpaz (changes requested):
One big thing still missing is tests. I don't think a lot of the new code is covered by existing tests. We should add tests that cover `Job` new methods.

anpaz (approved):
Still missing tests, but since this is merging into the feature branch I'm ok to merge it as it is.

azure-quantum/azure/quantum/optimization/solvers.py
guenp: Thanks for noticing! I will do that in the next PR.

azure-quantum/azure/quantum/optimization/solvers.py
guenp: I'm not sure if `to_blob` is the right place for that data. All that method does is convert a payload to a `bytes` object.
I think that the content type, encoding and input data type should all be properties of the `Problem`.

PR #86, opened 2021-07-09
Update third party notice

anjbur:


NOTICE
```diff
@@ -1,30 +1,3137 @@
...
-USA
 
-Notwithstanding any other terms, you may reverse engineer this software to the
-extent required to debug changes to any libraries licensed under the GNU Lesser
-General Public License.
+Copyright Jason R. Coombs
+
+MIT License
+
+Copyright (c) <year> <copyright holders>
```
guenp, line 3119: Should this line be filled out and/or removed?

guenp (approved):

NOTICE
anjbur: good catch! Taking a look at the license file in this component's repo, I think this is covered by line 3115. I went ahead and deleted this line since it's a duplicate.

PR #89, opened 2021-07-15
Adding optimization support for grouped cost function terms

eionblanc:
This PR enables the QDK to create and run Problems with cost functions that include squared linear combination (SLC) terms, which bear form C(w_1 x_1 + ... + w_d x_d)^2. Native support for these terms in factored form (that is, without having to expand them through multiplication) dramatically reduces input file sizes and quickens the solution computation. SLC terms usually result from feasibility constraints on optimization problems and thus crop up in problems such as vehicle routing, portfolio management, and job shop scheduling; efficient handling of SLC terms has been requested by multiple Azure Quantum clients needing to solve these problems.

With an eye for further extensions, the main contribution from enabling SLC term support is the more general architecture supporting the notion of _grouped terms_, one (and currently the only nontrivial) type of which is the SLC term. A grouped term contains a list of monomial terms which comprise the argument of a mathematical function (in the case of SLC terms, the squaring function), as well as a type (encoding the function at hand) and a lead coefficient.

The main technical features of this PR include:
- generalization of the Term class into classes for monomial terms and new grouped terms (including SLC)
- support for grouped term inputs in problem and streaming problem interfaces
- inclusion of unit tests for new features


adelebai:
converting to draft since feature is not complete yet.

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -124,12 +127,46 @@ def add_term(self, c: Union[int, float], indices: List[int]):
         self.terms.append(Term(indices=indices, c=c))
         self.uploaded_blob_uri = None
 
-    def add_terms(self, terms: List[Term]):
-        """Adds a list of terms to the `Problem` representation
+    def add_terms(self, terms: List[Term],
+                  type: str = None, c: Union[int, float] = 1):
```
adelebai, line 131: consider creating an enum

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -124,12 +127,46 @@ def add_term(self, c: Union[int, float], indices: List[int]):
...
+        :param type: Type of grouped term being added, if applicable
+        :param c: Weight of grouped term, if applicable
+        """
+        if type:
+            # Grouped term
+            if type == 'na':
+                gtype = GroupType.combination
+            elif type == 'slc':
+                gtype = GroupType.squared_linear_combination
+            else:
```
adelebai, line 145: wont be needed if you use enum for the type

azure-quantum/azure/quantum/optimization/problem.py
adelebai: I notice you already have GroupType - can we use that as an optional input? i.e. Optional[GroupType]

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -124,12 +127,46 @@ def add_term(self, c: Union[int, float], indices: List[int]):
...
+        to the `Problem` representation
 
         :param terms: The list of terms to add to the problem
+        :param type: Type of grouped term being added, if applicable
+        :param c: Weight of grouped term, if applicable
+        """
+        if type:
+            # Grouped term
+            if type == 'na':
+                gtype = GroupType.combination
```
adelebai, line 142: what's the difference between this and no type?

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -124,12 +127,46 @@ def add_term(self, c: Union[int, float], indices: List[int]):
...
+            else:
+                raise Exception("Unsupported group type {}.".format(type))
+            self.terms.append(GroupedTerm(gtype, terms, c=c))
+        else:
+            # List of ungrouped monomial terms
+            self.terms += terms
+        self.uploaded_blob_uri = None
+    
+    def add_slc_term(self,
+                     terms: List[Tuple[Union[int, float], Union[int, Type[None]]]],
```
adelebai, line 154: I remember we discussed to allow both a list of terms as well as the list of tuples. 
Also i think  Union[int, Type[None]] can be replaced with Optional[int]

azure-quantum/azure/quantum/optimization/problem.py
eionblanc: I allowed for type to be either GroupType or str to leverage the specificity one with the flexibility of the other.

azure-quantum/azure/quantum/optimization/problem.py
eionblanc: When type is entered as the GroupType, the string comparisons are now skipped over.

azure-quantum/azure/quantum/optimization/problem.py
eionblanc: No type means that the list of terms are added without being grouped into a face, which means that (if there are no other grouped terms), the model used for that cost function will be the more basic ones (without faces or SLC terms, etc). Having type 'na' is one way to trigger the SLC models even with a cost function that doesn't need it (and is built into the SLC models anyway, so I think it's good practice to have this accessible via QDK).

azure-quantum/azure/quantum/optimization/problem.py
eionblanc: The add_slc_term accepts either a list of tuples or a list of Term objects (new); the add_terms function accepts a list of Term objects and optionally groups them according to a type specification (including SLC).

azure-quantum/azure/quantum/optimization/problem.py
adelebai: Why allow for strings in the first place when there's an enum available?

azure-quantum/azure/quantum/optimization/streaming_problem.py
```diff
@@ -153,6 +184,43 @@ def add_terms(self, terms: List[Term]):
             min_coupling = min(term_couplings)
             self.__n_couplers += sum(term_couplings)
             self.stats["num_terms"] += len(terms)
+
+            for term in terms:
```
adelebai, line 188: is this supposed to override the above? (lines 182-186)

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,141 @@ def reduce_by_variable_state(
...
         return Term(indices=new_ids, c=new_c)
 
-    def __repr__(self):
-        return str(self.__dict__)
 
-    def __eq__(self, other):
-        if isinstance(other, self.__class__):
-            return self.__dict__ == other.__dict__
+class GroupType(str, Enum):
+    combination = "na"
```
adelebai, line 199: if I understand, a GroupTerm with type "na" is the same as a regular list of terms? So we are introducing another way to add regular terms?

adelebai: I think it's mostly harmless to add this, but it's more functionality that we didn't need.

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,141 @@ def reduce_by_variable_state(
...
+                    )
+                elif len(term.ids) == 1:
+                    # Linear term
+                    id = term.ids[0]
+                else:
+                    # Constant term
+                    id = -1
+                if id in seen:
+                    raise ValueError(
+                        "Error - uncombined like terms in Term with type {}".format(self.type)
```
adelebai, line 236: "multiple terms with the same variable must be combined"

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -124,12 +127,53 @@ def add_term(self, c: Union[int, float], indices: List[int]):
         self.terms.append(Term(indices=indices, c=c))
         self.uploaded_blob_uri = None
 
-    def add_terms(self, terms: List[Term]):
-        """Adds a list of terms to the `Problem` representation
+    def add_terms(self, terms: List[Term],
+                  type: Union[Optional[GroupType], str] = None, c: Union[int, float] = 1):
```
adelebai: in my opinion we have 2 ways of handling this enum if we want to keep the "na" group type. We can remove the Optional, and set the default value to GroupType.combination. 
If we want to keep the None, then I think we get rid of the GroupType.combination entirely. I don't see where the requirement is to support the existing monomials but as a nested group.

adelebai (changes requested):
I would consider getting rid of GroupTerms supporting a regular list of monomials entirely, when there's already the existing mechanism to add ungrouped terms.

azure-quantum/azure/quantum/optimization/streaming_problem.py
eionblanc: Yes, good catch; thank you.

azure-quantum/azure/quantum/optimization/term.py
eionblanc: Keeping the `combination` type but re-routing the interface of adding a grouped combination of terms to an ungrouped list of monomials as we discussed

azure-quantum/azure/quantum/optimization/problem.py
eionblanc: Implemented the first option

eionblanc:
> I would consider getting rid of GroupTerms supporting a regular list of monomials entirely, when there's already the existing mechanism to add ungrouped terms.

The add_terms function now routes away from instantiating a GroupedTerm with type combination; similarly, if a GroupTerm with type combination is initialized, the user receives a warning indicating to use a list of ungrouped monomials instead.

adelebai (approved):

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -96,15 +99,57 @@ def __init__(
             )
         self.c = coeff
 
-        self.ids = indices
-
     def to_dict(self):
         return self.__dict__
 
     @staticmethod
     def from_dict(obj):
```
guenp, line 106:
```suggestion
    @classmethod
    def from_dict(cls, obj: dict):
```

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -96,15 +99,57 @@ def __init__(
...
-        return Term(indices=obj["ids"], c=obj["c"])
+        return GenTerm(c=obj["c"])
```
guenp, line 107:
```suggestion
        return cls(c=obj.get("c"))
```

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -96,15 +99,57 @@ def __init__(
...
+        indices: List[int] = None,
+        w: Optional[WArray] = None,
+        c: Optional[WArray] = None,
+    ):
+        GenTerm.__init__(self, w=w, c=c)
+        self.ids = indices
+    
+    @staticmethod
+    def from_dict(obj):
+        return Term(indices=obj["ids"], c=obj["c"])
```
guenp, line 151:
```suggestion
    @classmethod
    def from_dict(cls, obj: dict):
        return cls(indices=obj.get("ids"), c=obj.get("c"))
```

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,146 @@ def reduce_by_variable_state(
...
+        
+        try:
+            terms = [Term.from_dict(term_dict) for term_dict in obj["terms"]]
+        except:
+            print(
+                "Error - grouped list of terms missing or errant."
+            )
+            raise
+
+        return GroupedTerm(type=type, terms=terms, c=obj["c"])
```
guenp, line 277:
```suggestion
    @classmethod
    def from_dict(cls, obj: dict):
        """
        Create GroupedTerm from dictionary with keys "type", "terms" and "c"
        """
        if obj["type"] == "na":
            type = GroupType.combination
        elif obj["type"] == "slc":
            type = GroupType.squared_linear_combination
        else:
            print(
                "Error - unknown grouped term type {0}".format(
                    obj["type"]
                )
            )
            raise
        
        try:
            terms = [Term.from_dict(term_dict) for term_dict in obj["terms"]]
        except:
            print(
                "Error - grouped list of terms missing or errant."
            )
            raise

        return cls(type=type, terms=terms, c=obj["c"])
```

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,146 @@ def reduce_by_variable_state(
...
+    
+    def evaluate(self, configuration: Dict[int, int]) -> float:
+        """Given a variable configuration, evaluate the value of the grouped term.
+        :param configuration:
+            The dictionary of variable ids to their assigned value
+        """
+        if self.type is GroupType.combination:
+            combination_eval = 0.0
+            for monomial in self.terms:
```
guenp, line 286: Here I am also confused by the usage of the word `monomial`, why not simply use `for term in self.terms`?

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,146 @@ def reduce_by_variable_state(
...
+                "Error - evaluate not handled for GroupType {0}".format(
+                    self.type
+                )
+            )
+            raise
+        
+        return eval
+    
+    def reduce_by_variable_state(
+        self, fixed_variables: Dict[int, int]
```
guenp, line 305:
```suggestion
        self,
        fixed_variables: Dict[int, int]
```

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,146 @@ def reduce_by_variable_state(
...
+    ) -> Optional[GenTerm]:
+        """Given some fixed variable states,
+            transform the existing grouped term into new term.
+        Returns None if the new term is effectively 0
+        :param fixed_variables:
+            The dictionary of variable ids and their fixed state
```
guenp, line 311: Perhaps add some info here on the expected keys

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,146 @@ def reduce_by_variable_state(
...
+            return None
+
+        # GroupType simplifications when new_terms has a single element
+        # Further simplifications require knowledge of binary vs. spin setting
+        if len(new_terms) == 1:
+            term = new_terms[0]
+            if self.type is GroupType.combination:
+                return Term(indices=term.ids, c=self.c * term.c)
+            elif self.type is GroupType.squared_linear_combination:
+                #return Term(indices=term.ids + term.ids, c=self.c * term.c**2)
```
guenp, line 333:
```suggestion
```

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,146 @@ def reduce_by_variable_state(
...
+                if len(term.ids) == 0:
+                    return Term(indices=term.ids, c=self.c * term.c**2)
```
guenp, line 335: what does this equation `self.c * term.c **2` mean? could you add an inline comment to clarify?

guenp (changes requested):
First pass, I think it looks like a great addition so far. Some documentation or examples on the changes would be helpful.

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -290,15 +335,21 @@ def download(self, workspace: "Workspace"):
         contents = download_blob(blob.url)
         return Problem.deserialize(contents, self.name)
 
-    def get_terms(self, id: int) -> List[Term]:
+    def get_terms(self, id:nint) -> List[GenTerm]:
```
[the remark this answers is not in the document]
eionblanc: This was a typo with the 'n'; thank you for catching it!

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -96,15 +99,57 @@ def __init__(
...
-
     def to_dict(self):
         return self.__dict__
 
     @staticmethod
     def from_dict(obj):
-        return Term(indices=obj["ids"], c=obj["c"])
+        return GenTerm(c=obj["c"])
+
+    def evaluate(self, configuration: Dict[int, int]) -> float:
```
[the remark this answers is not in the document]
eionblanc: Thank you; I included these to be consistent with the arguments needed downstream but see that this works better instead.

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -96,15 +99,57 @@ def __init__(
...
 
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            return self.__dict__ == other.__dict__
+        else:
+            return False
+
+class Term(GenTerm):
+    """
+    Monomial term class
```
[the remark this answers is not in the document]
eionblanc: I rephrased this a bit to try to meet the best of both worlds; I think emphasizing the distinction between monomial/single terms and grouped terms is important since grouped terms (like SLC) are considered single, non-monomial terms.

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -149,11 +194,146 @@ def reduce_by_variable_state(
...
+        GenTerm.__init__(self, w=w, c=c)
+        self.type = type
+        self.terms = terms
+        self.validate()
+    
+    def validate(self):
+        if self.type is GroupType.combination:
+            raise ValueError(
+                "Error - type {} GroupTerm is not enabled;"
+                + "formulate list of ungrouped Term monomials instead.".format(self.type)
```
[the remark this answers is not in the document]
eionblanc: I rephrased this accordingly; including monomial again was to emphasize that Term objects are not grouped terms.

azure-quantum/azure/quantum/optimization/term.py
eionblanc: Emphasizing that the terms composing grouped terms must be monomials is useful but not necessary since it is structurally enforced; I renamed it to `term `for consistency.

guenp (approved):

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -69,10 +70,12 @@ def _convert_if_numpy_type(param: WArray):
...
 
-class Term:
+class GenTerm:
+    """
+    Generic term class; rarely to be directly initialized.
+    """
     def __init__(
         self,
-        indices: List[int] = None,
         w: Optional[WArray] = None,
```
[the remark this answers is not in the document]
frtibble: Let's discuss this in standup today @eionblanc

frtibble (approved):

adelebai:
As discussed yesterday - we want to introduce a new model type before we merge this.

adelebai:
> As discussed yesterday - we want to introduce a new model type before we merge this.

In addition, we should add a temporary check in solver.optimize to reject the job for the wrong solver type (if grouped terms are used)... now that we don't need to iterate through all the terms

adelebai (changes requested):
pubo/ising grouped type, and some input checking to reject solvers that are not PA/SSMC

azure-quantum/azure/quantum/optimization/term.py
```diff
@@ -69,10 +71,12 @@ def _convert_if_numpy_type(param: WArray):
...
 
-class Term:
+class TermBase(ABC):
+    """
+    Term base class; this class is not directly initialized
+    """
     def __init__(
         self,
-        indices: List[int] = None,
         w: Optional[WArray] = None,
```
adelebai, line 80: on second glance, this is not being used for serialization so we can remove the need for it in TermBase, but keep it in Term for now.

adelebai: If we want to print that user friendly message for "w" or "c", we can print it in "Term".

adelebai:
This is a request from Andres - please elaborate on the feature a bit more in the PR description, customer impact, usage and what the use case is.

adelebai (approved):

azure-quantum/azure/quantum/optimization/streaming_problem.py
```diff
@@ -148,11 +199,43 @@ def add_terms(self, terms: List[Term]):
...
 
-            term_couplings = [len(term.ids) for term in terms]
-            max_coupling = max(term_couplings)
-            min_coupling = min(term_couplings)
-            self.__n_couplers += sum(term_couplings)
-            self.stats["num_terms"] += len(terms)
+            for term in terms:
+                if isinstance(term, Term):
+                    n = len(term.ids)
+                    max_coupling = max(max_coupling, n)
```
guenp, line 205: This is throwing an error in the live tests because `max_coupling` is not defined

guenp: I was able to fix this locally by adding the following lines above ln 202:
```python
term_couplings = [len(term.ids) for term in terms]
max_coupling = max(term_couplings)
min_coupling = min(term_couplings)
```

PR #93, opened 2021-07-19
No print on wait

tonybaloney:
The `Job.wait_until_completed()` method prints `.` to the console.

Users can fetch the status and the task is blocking. There's no way to disable the print and this is the only method to push data to the console.

### DIFF
--- a/azure-quantum/azure/quantum/job.py
+++ b/azure-quantum/azure/quantum/job.py
@@ -51,7 +51,7 @@ class Job:
             or self.details.status == "Cancelled"
         )
 
-    def wait_until_completed(self, max_poll_wait_secs=30):
+    def wait_until_completed(self, max_poll_wait_secs=30, print_progress=True):
         """Keeps refreshing the Job's details
         until it reaches a finished status."""
         self.refresh()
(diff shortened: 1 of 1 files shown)


vxfield:
Thanks, @tonybaloney.
I think this was implemented to get some visual progress during interactive scenarios, like in a Jupyter Notebook.

@anpaz, @guenp: What do you think about this scenario? Maybe we can add an optional parameter to disable the print(".") (which will need to be propagated to `get_results`)?

azure-quantum/azure/quantum/job.py
```diff
@@ -61,7 +61,6 @@ def wait_until_completed(self, max_poll_wait_secs=30):
                 f"Waiting for job {self.id},"
                 + f"it is in status '{self.details.status}'"
             )
-            print(".", end="", flush=True)
             time.sleep(poll_wait)
```
guenp, line 64:
```suggestion
            if print_progress:
                print(".", end="", flush=True)
            time.sleep(poll_wait)
```

guenp (changes requested):
I think it's a good idea to disable the print, but to delete it would cause a change in behavior that may confuse some users. Let's add an optional input argument `print_progress` that defaults to `True` to `wait_ujntil_completed`.

tonybaloney:
@guenp changes made

anpaz (commented):
lgtm.

guenp (approved):

Issue #94, opened 2021-07-19
async operations not exposed via workspace API


tonybaloney:
The azure SDK has async methods for all of the operations for the Quantum API. 

It would be really handy to have async methods as this is the type of work you'd want to await upon.

I'll raise a PR

guenp:
Thanks @toneybaloney for your suggestion! Could you share with us your use case or a scenario in which you'd use async i/o with the Azure Quantum service?

PR #95, opened 2021-07-19
Add an async workspace API

tonybaloney:
Implements #94 

- Async interfaces are within an `aio` submodule to match the Python SDK pattern
- Extends the base workspace class to minimise the amount of duplicate code
- Implements an async job class
- Had to expose async solvers to inherit from an async solver class. Used the synchronous solver as a base to avoid duplication

The biggest amount of duplication here is because of the `_authentication` submodule and the default credential class. All of the Python SDKs would expect `get_token` to be async, which comes from different interfaces. This could be refactored.

Please let me know your thoughts


tonybaloney:
This is the modified version of the shipping example I'm using to test: 

```python
#!/usr/bin/env python
# coding: utf-8
# Copyright (c) Microsoft Corporation.
# Licensed under the MIT License.

import asyncio
import os
from typing import List, Sequence

# Instantiate Workspace object which allows you to connect to the Workspace you've previously deployed in Azure.
# Be sure to fill in the settings below which can be retrieved by running 'az quantum workspace show' in the terminal.
from azure.quantum import AsyncWorkspace
from azure.quantum.optimization import (ParallelTempering, Problem,
                                        ProblemType, Term, AsyncSolver)
from azure.identity.aio import DefaultAzureCredential
from rich.console import Console
from rich.table import Table

import numpy as np

NUM_CONTAINERS = 100
CONTAINER_MIN_WEIGHT = 1
CONTAINER_MAX_WEIGHT = 50

# Copy the settings for your workspace below
workspace = AsyncWorkspace(
    credential=DefaultAzureCredential(),
    resource_id = os.getenv('AZURE_RESOURCE_ID'), # add the Resource ID of your Azure Quantum workspace
    location = os.getenv('AZURE_LOCATION', 'westus')     # add the location of your Azure Quantum workspace (e.g. "westus")
)


def create_problem_for_container_weights(container_weights: Sequence[int]) -> Problem:
    terms: List[Term] = []

    # Expand the squared summation
    for i in range(len(container_weights)):
        for j in range(len(container_weights)):
            if i == j:
                # Skip the terms where i == j as they form constant terms in an Ising problem and can be disregarded:
                # w_i∗w_j∗x_i∗x_j = w_i​*w_j∗(x_i)^2 = w_i∗w_j​​
                # for x_i = x_j, x_i ∈ {1, -1}
                continue

            terms.append(
                Term(
                    c = container_weights[i] * container_weights[j],
                    indices = [i, j]
                )
            )

    # Return an Ising-type problem
    return Problem(name="Ship Sample Problem", problem_type=ProblemType.ising, terms=terms)

# This array contains a list of the weights of the containers
rng = np.random.default_rng()
container_weights = rng.integers(low=CONTAINER_MIN_WEIGHT, high=CONTAINER_MAX_WEIGHT, size=NUM_CONTAINERS)

# Create a problem for the list of containers:
problem1 = create_problem_for_container_weights(container_weights)

class AsyncParallelTempering(ParallelTempering, AsyncSolver):
    pass

# Instantiate a solver to solve the problem.
parallel_solver = AsyncParallelTempering(workspace, timeout=100)

# Print out a summary of the results:
def print_result_summary(result, console):
    table = Table(title="Container Allocations")

    table.add_column("Container", justify="right", style="cyan", no_wrap=True)
    table.add_column("Weight", style="magenta")
    table.add_column("Ship", justify="right", style="green")

    # Print a summary of the result
    ship_a_weight = 0
    ship_b_weight = 0
    for container in result['configuration']:
        container_assignment = result['configuration'][container]
        container_weight = container_weights[int(container)]
        ship = ''
        if container_assignment == 1:
            ship = 'A'
            ship_a_weight += container_weight
        else:
            ship = 'B'
            ship_b_weight += container_weight

        table.add_row(container, str(container_weight), ship)

    console.print(table)
    print(f'\nTotal weights: \n\tShip A: {ship_a_weight} tonnes \n\tShip B: {ship_b_weight} tonnes\n')


# We can reduce the number of terms by removing duplicates (see associated Jupyter notebook for details)
# In code, this means a small modification to the create_problem_for_container_weights function:
def create_simplified_problem_for_container_weights(container_weights: List[int]) -> Problem:
    terms: List[Term] = []

    # Expand the squared summation
    for i in range(len(container_weights)-1):
        for j in range(i+1, len(container_weights)):
            terms.append(
                Term(
                    c = container_weights[i] * container_weights[j],
                    indices = [i, j]
                )
            )

    # Return an Ising-type problem
    return Problem(name="Ship Sample Problem (Simplified)", problem_type=ProblemType.ising, terms=terms)

# Check that this creates a smaller problem
# Create the simplified problem
problem2 = create_simplified_problem_for_container_weights(container_weights)


async def submit_problems():
    # Schedule three calls *concurrently*:
    results = await asyncio.gather(
        parallel_solver.optimize(problem1),
        parallel_solver.optimize(problem2)
    )

    console = Console()
    print_result_summary(results[0], console)
    print_result_summary(results[1], console)


asyncio.run(submit_problems())
```

guenp:
@tonybaloney I just tried out this feature using the sample copied above but unfortunately it looks like the credentials are incompatible with `asyncio`:
```
adal.adal_error.AdalError: Get Token request returned http error: 400 and server response: {"error":"interaction_required","error_description":"AADSTS70043: The refresh token has expired or is invalid due to sign-in frequency checks by conditional access. The token was issued on 2021-06-22T07:28:32.1570000Z and the maximum allowed lifetime for this request is 43200.\r\nTrace ID: 2f1a2f47-9079-4661-8056-93164e4f6e00\r\nCorrelation ID: 86c30975-a8f9-4b6c-bbd8-e537d318a95a\r\nTimestamp: 2021-07-26 17:09:17Z","error_codes":[70043],"timestamp":"2021-07-26 17:09:17Z","trace_id":"2f1a2f47-9079-4661-8056-93164e4f6e00","correlation_id":"86c30975-a8f9-4b6c-bbd8-e537d318a95a","suberror":"token_expired"}
To open an issue, please run: 'az feedback'

	AzurePowerShellCredential
	InteractiveBrowserCredential: object AccessToken can't be used in 'await' expression
	DeviceCodeCredential: object AccessToken can't be used in 'await' expression
```
Am I missing a step?

tonybaloney:
See https://github.com/Azure/azure-sdk-for-python/issues/19943

azure-quantum/azure/quantum/optimization/solvers.py
```diff
@@ -78,6 +78,46 @@ def __init__(
     SWEEPS_WARNING = 10000
     TIMEOUT_WARNING = 600
 
+    def _jobdetails_from_problem(
```
guenp, line 81: This is a nice refactor; however, we currently have a feature branch open where we've refactored this in a different way (see #108), I suggest let's wait until that is merged into `main` and pull in the changes here

guenp: It is now refactored, see here for the `Solver.submit` method:
https://github.com/microsoft/qdk-python/blob/1f2607e0c22a27417635a98991fbd9f803dc34d2/azure-quantum/azure/quantum/optimization/solvers.py#L76

vxfield:
@tonybaloney, could you create a new PR to merge it in the branch `feature/aio`?
This way Guen and I can apply some final touches before merging this to `main`.

guenp:
@tonybaloney, I opened a PR here to update this branch with the latest changes from `main`.
There is still some more work to do after this (e.g. TargetFactory and _TokenCredential) but we can work on that after this is merged into the feature branch.

guenp (approved):
Thank you for the awesome contribution! 🙌

PR #98, opened 2021-07-19
azure-quantum unit test recording infrastructure fixes

### LINKED ISSUE #92: Playback recording does not record fetched job results, stores secrets
When using playback recording, the fetched job results are not recorded and returned during playback. Also, it looks like the recording currently contains secrets, and not all sub ID occurrences are replaced by the dummy job ID (`0000-...`).

@vxfield investigated and the issue is twofold:

1. The `content-range`  header is missing from the `CustomRecordingProcessor` allow list
2. Interactive credentials adds additional secrets.

The proposed solution is to add the missing header, add sanitization steps for the output YAML in the `regex_replacer`, and add a new `InteractiveAccessTokenReplacer` class that replaces any tokens generated by the interactive credential.

vxfield:
Fixes #92 by:
1) Adding `content-disposition` and `content-range` HTTP Headers that are used by Storage SDK blob download in the test recording allow-list
2) Introducing a `InteractiveAccessTokenReplacer` to sanitize interactive credentials from the test recordings
3) Sanitize HTTP Headers with the same regexes that we use to sanitize the HTTP URI and Body

### DIFF
--- a/azure-quantum/tests/unit/common.py
+++ b/azure-quantum/tests/unit/common.py
@@ -101,6 +101,9 @@ class QuantumTestBase(ReplayableTest):
             r"jobs/([a-f0-9]+[-]){4}[a-f0-9]+", "jobs/" + ZERO_UID
         )
         regex_replacer.register_regex(
+            r"job-([a-f0-9]+[-]){4}[a-f0-9]+", "job-" + ZERO_UID
+        )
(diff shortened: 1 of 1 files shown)


guenp (approved):

guenp:
@vxfield it looks like there's a snafu in the build system, I am OK to merge this as is

PR #99, opened 2021-07-19
`azure-quantum` as the default conda environment name

vxfield:
Make `azure-quantum` the default conda environment name and consistent across readme instructions and the `environment.yml` file.


guenp:
We also need to change this line (from https://github.com/microsoft/qdk-python/blob/main/build/test.ps1#L25)
```
$EnvNames = $PackageDirs | ForEach-Object {$_.replace("-", "")}
```

into:
```
$EnvNames = $PackageDirs
```

vxfield:
Thanks. Just pushed a commit with the suggested change.

guenp:
It looks like it's still not working: https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=29765&view=logs&j=20fe306e-41c1-59b2-a28d-701712929519&t=eedfd4a5-c0bd-59b7-5c8c-a6d741cad3fe&l=15

I think the reason why we went with `azurequantum` as a name for the env was because of this issue. I think the dash still gets dropped when creating the environment.

vxfield:
Conda allows dashes in the env names, so I'm not sure where it's getting removed.
I'll take another look at what's going on...

guenp:
Unfortunately it didn't get fixed. We are currently using only environment names without hyphens.. I'm still not sure where the hyphens are getting dropped.

PR #100, opened 2021-07-19
importing models from _client

adelebai:
Expose some client models for the user (e.g. JobStatus)

### DIFF
--- a/azure-quantum/azure/quantum/__init__.py
+++ b/azure-quantum/azure/quantum/__init__.py
@@ -8,5 +8,7 @@ from .version import __version__
 from .job import *
 from .workspace import *
 
+from ._client.models import *
+
 logger = logging.getLogger(__name__)
 logger.info(f"version: {__version__}")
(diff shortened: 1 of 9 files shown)


azure-quantum/azure/quantum/__init__.py
```diff
@@ -8,5 +8,7 @@
 from .job import *
 from .workspace import *
 
+from ._client.models import *
```
guenp, line 11:
```suggestion
from ._client.models._quantum_client_enums import JobStatus
```

guenp (approved):
I would suggest to use a direct import from `._client.models._quantum_client_enums` instead of the wildcard `from foo import *`; the latter also makes other models available that we want to keep private.

azure-quantum/azure/quantum/__init__.py
adelebai: I think it's ._client.models import JobStatus (there's no _quantum_client_enums module). 
Let me make the new suggestion

guenp: Yes, it is also available via _client.models, but the class is defined here: https://github.com/microsoft/qdk-python/blob/main/azure-quantum/azure/quantum/_client/models/_quantum_client_enums.py#L36

adelebai: ok changed

PR #101, opened 2021-07-20
Add Honeywell provider with QASM passthrough

guenp:
Base should be changed to `feature/3rdparty-integration` after #79 is merged


azure-quantum/tests/unit/test_target.py
```diff
@@ -79,7 +80,72 @@ def _test_job_submit_ionq(self, num_shots):
...
+
+            # Make sure the job is completed before fetching the results
+            # playback currently does not work for repeated calls
+            if not self.is_playback:
+                self.assertEqual(False, job.has_completed())
+                if self.in_recording:
+                    import time
+                    time.sleep(3)
+                job.refresh()
+                job.wait_until_completed()
```
anpaz, line 144: I don't think we should wait for completion; unfortunately honeywell system is not mature enough that they can guarantee that their targets will have 99.99% availability, not even their api validator.

anpaz (approved):
see comments, though.

azure-quantum/azure/quantum/target/honeywell.py
```diff
@@ -0,0 +1,56 @@
...
+from azure.quantum.workspace import Workspace
+
+
+class Honeywell(Target):
+    """Honeywell target."""
+
+    def __init__(
+        self,
+        workspace: Workspace,
+        target: str = "honeywell.hqs-lt-s1-apival",
```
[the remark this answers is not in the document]
guenp: I set the `ionq.simulator` target as the default for IonQ as well so I thought it would be good to be consistent. Do you think this should simply have no default value?
Nonetheless I think we still need a `Workspace.list_providers()` method that lists all the accessible providers and targets.

azure-quantum/tests/unit/test_target.py
```diff
@@ -79,7 +80,72 @@ def _test_job_submit_ionq(self, num_shots):
...
+    def test_job_submit_honeywell(self):
+
+        with unittest.mock.patch.object(
+            Job,
+            self.mock_create_job_id_name,
+            return_value=self.get_test_job_id(),
+        ):
+            workspace = self.create_workspace()
+            circuit = self._teleport()
+            target = Honeywell(workspace=workspace)
```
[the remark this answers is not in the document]
guenp: Do we want the end to end tests to run on hardware? For IonQ it is currently running against the simulator.

azure-quantum/tests/unit/test_target.py
guenp: This is meant for when we are recording playback for the API. In that case, we do want to wait until it is completed.
If we are not recording and running this in CI, we can set a timeout and check the status of the job later.

guenp: I'm merging this to the feature branch for now, let's loop back on the e2e scenario and how to test it when we are ready.

PR #107, opened 2021-07-27
Add timeout for wait_until_completed, make HW test more robust against downtime

guenp:
Adds a `timeout` argument to `Job.wait_until_completed` method, implements timeout on Honeywell service, catch `HttpResponseError` to mitigate missing Honeywell provider in E2E tests container

### DIFF
--- a/azure-quantum/azure/quantum/job/job.py
+++ b/azure-quantum/azure/quantum/job/job.py
@@ -56,7 +56,7 @@ class Job(BaseJob, FilteredJob):
             or self.details.status == "Cancelled"
         )
 
-    def wait_until_completed(self, max_poll_wait_secs=30, timeout_secs=DEFAULT_TIMEOUT) -> None:
+    def wait_until_completed(self, max_poll_wait_secs=30, timeout_secs=None) -> None:
         """Keeps refreshing the Job's details
         until it reaches a finished status.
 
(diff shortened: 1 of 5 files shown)


azure-quantum/azure/quantum/job/job.py
```diff
@@ -56,18 +56,30 @@ def has_completed(self) -> bool:
             or self.details.status == "Cancelled"
         )
 
-    def wait_until_completed(self, max_poll_wait_secs=30) -> None:
+    def wait_until_completed(self, max_poll_wait_secs=30, timeout_secs=None) -> None:
```
anpaz, line 59: I'm not sure, do we want to give some default timeout, say 5 mins?

azure-quantum/tests/unit/test_target.py
```diff
@@ -131,21 +133,29 @@ def test_job_submit_honeywell(self):
...
+                # Make sure the job is completed before fetching the results
+                # playback currently does not work for repeated calls
+                if not self.is_playback:
+                    self.assertEqual(False, job.has_completed())
+                    if self.in_recording:
+                        import time
+                        time.sleep(3)
+                    job.refresh()
+                    try:
+                        job.wait_until_completed(timeout=5*60) # Set a timeout for Honeywell recording
```
anpaz, line 150: Considering this might fail most of the time, I would set the timeout to only 1 minute. The job itself should take no time, so not much point on waiting for Honeywell to come online if it's not available right now.

anpaz (approved):

azure-quantum/azure/quantum/job/job.py
guenp: That would be a breaking change, let's double check with the team if this is a desired change for QIO jobs.

azure-quantum/tests/unit/test_target.py
guenp: Sounds good

azure-quantum/azure/quantum/job/job.py
guenp: Set default to 180s

PR #110, opened 2021-07-28
Add note on autogenerated code for contributors

guenp:
Added a note for contributors on autogenerated code.


anpaz (approved):

vxfield (approved):
Have a suggestion to make it clear that we won't accept changes to `_client` code.

PR #112, opened 2021-07-28
Add get_targets and get_quotas to Workspace

guenp:
Add `Workspace.get_targets` and `Workspace.get_job_quotas`
Depends on #111

### DIFF
--- a/azure-quantum/azure/quantum/workspace.py
+++ b/azure-quantum/azure/quantum/workspace.py
@@ -7,7 +7,7 @@ import logging
 import os
 import re
 
-from typing import Dict, List, Optional
+from typing import Any, Dict, List, Optional
 from deprecated import deprecated
 
 # Temporarily replacing the DefaultAzureCredential with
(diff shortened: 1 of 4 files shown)


azure-quantum/azure/quantum/workspace.py
```diff
@@ -257,6 +257,27 @@ def list_jobs(
...
+            ] for provider in self._client.providers.get_status()
+        }
+
+    def get_job_quotas(self) -> List[Dict[str, Any]]:
+        """Get a list of job quotas for the given workspace.
+
+        :return: Job quotas
+        :rtype: List[Dict[str, Any]]
+        """
+        return [q.as_dict() for q in self._client.quotas.list()]
```
anpaz, line 279: `self._client.quotas` should be wrapped in its own `_get_quotas_client`

anpaz (approved):
Just waiting on release build to make sure Live tests work well.

azure-quantum/azure/quantum/workspace.py
```diff
@@ -257,6 +257,27 @@ def list_jobs(
...
+        :return: Targets, keyed by provider IDs
+        :rtype: Dict[str, List[str]]
+        """
+        return {
+            provider.id: [
+                target.id for target in provider.targets
+            ] for provider in self._client.providers.get_status()
+        }
+
+    def get_job_quotas(self) -> List[Dict[str, Any]]:
```
anpaz, line 273: job_quotas, seems to refer as the quotas of a single job, maybe just keep it `quotas`?

azure-quantum/azure/quantum/workspace.py
guenp: Sure, good idea to keep it consist

PR #113, opened 2021-07-29
Change env variable names to match other QDK Live tests.

anpaz:
Just change the name of environment variables to match those of live tests (i.e. use the AZUREQUANTUM prefix to avoid conflicts with other variables).


guenp (commented):
I would suggest to use AZURE_QUANTUM instead of AZUREQUANTUM as the prefix because they are separate words.

anpaz:
I don't disagree, but this the name we already use and it would require changes in multiple repos. The aim is just to bring this one in-sync with the rest.

guenp (approved):

PR #114, opened 2021-07-29
Add github action to trigger qdk-sync pipeline

anpaz:


guenp (approved):
Thanks for catching this!

PR #119, opened 2021-08-03
Add azure-quantum, azure-cli dependencies

guenp:
Add dependencies to `qdk`.

### DIFF
--- a/qdk/setup.py
+++ b/qdk/setup.py
@@ -39,6 +39,8 @@ setuptools.setup(
         'ruamel-yaml',
         'basis_set_exchange',
         'jupyter_nbextensions_configurator',
-        'pygments>=2.7.4'
+        'pygments>=2.7.4',
+        'azure-quantum==' + version,
+        'azure-cli>=2.23.0',
     ]
 )


qdk/setup.py
```diff
@@ -39,6 +39,8 @@
         'ruamel-yaml',
         'basis_set_exchange',
         'jupyter_nbextensions_configurator',
-        'pygments>=2.7.4'
+        'pygments>=2.7.4',
+        'azure-quantum==' + version,
```
vxfield, line 43: This will only work if `azure-quantum` is built together with this package and is either published, available in a local package source, or previously installed.

For now, I think we should manually set the latest published version of `azure-quantum` or update the build scripts to build and install  `azure-quantum`.
Note that if we are going to move the code-base to `azure-sdk-for-python` repo, the later option will be more complicated because it will involve kicking-off another ADO pipeline, waiting for its results and grabbing the build artifacts.

vxfield (changes requested):

qdk/setup.py
guenp: I think I'll close this PR for now then, let's revisit this when we decide on moving the repo.

PR #120, opened 2021-08-04
Correct `to_blob()` rtype

tonybaloney:
This method would only ever return `bytes`, not `Tuple[bytes, str`


guenp (approved):
Nice catch, thanks so much!

PR #122, opened 2021-08-10
Minor fixes to grouped terms

adelebai:
Minor fixes to grouped terms.
Running live tests again to confirm this fixes the problem.

### DIFF
--- a/azure-quantum/tests/unit/test_job.py
+++ b/azure-quantum/tests/unit/test_job.py
@@ -127,13 +127,15 @@ class TestJob(QuantumTestBase):
         solver_type = functools.partial(microsoft.PopulationAnnealing, sweeps=200)
         solver_name = "PopulationAnnealing"
         self._test_job_submit(solver_name, solver_type)
-        self._test_job_submit(solver_name, solver_type, test_grouped=True)
+        # todo - re-enable once grouped model is live
+        #self._test_job_submit(solver_name, solver_type, test_grouped=True)
 
     def test_job_submit_microsoft_substochastic_monte_carlo(self):
         solver_type = functools.partial(microsoft.SubstochasticMonteCarlo, step_limit=280)
         solver_name = "SubstochasticMonteCarlo"
         self._test_job_submit(solver_name, solver_type)
-        self._test_job_submit(solver_name, solver_type, test_grouped=True)
+        # todo - re-enable once grouped model is live
+        #self._test_job_submit(solver_name, solver_type, test_grouped=True)
 
     def test_job_upload_and_run_solvers(self):
         problem_name = f'Test-problem-{datetime.now():"%Y%m%d-%H%M%S"}'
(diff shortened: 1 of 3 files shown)


azure-quantum/tests/unit/test_job.py
```diff
@@ -127,13 +127,15 @@ def test_job_submit_microsoft_population_annealing(self):
         solver_type = functools.partial(microsoft.PopulationAnnealing, sweeps=200)
         solver_name = "PopulationAnnealing"
         self._test_job_submit(solver_name, solver_type)
-        self._test_job_submit(solver_name, solver_type, test_grouped=True)
+        # todo - re-enable once grouped model is live
+        #self._test_job_submit(solver_name, solver_type, test_grouped=True)
```
guenp, line 131: I think we can leave this test live but instead just set `test_grouped` to `False`.
```suggestion
        self._test_job_submit(solver_name, solver_type, test_grouped=False)
```

azure-quantum/tests/unit/test_job.py
```diff
@@ -127,13 +127,15 @@ def test_job_submit_microsoft_population_annealing(self):
...
 
     def test_job_submit_microsoft_substochastic_monte_carlo(self):
         solver_type = functools.partial(microsoft.SubstochasticMonteCarlo, step_limit=280)
         solver_name = "SubstochasticMonteCarlo"
         self._test_job_submit(solver_name, solver_type)
-        self._test_job_submit(solver_name, solver_type, test_grouped=True)
+        # todo - re-enable once grouped model is live
+        #self._test_job_submit(solver_name, solver_type, test_grouped=True)
```
guenp, line 138:
```suggestion
        self._test_job_submit(solver_name, solver_type, test_grouped=False)
```

azure-quantum/tests/unit/test_target.py
```diff
@@ -148,7 +148,7 @@ def test_job_submit_honeywell(self):
                     self.assertEqual(False, job.has_completed())
                     try:
                         # Set a timeout for Honeywell recording
-                        job.wait_until_completed(timeout=60)
+                        job.wait_until_completed(timeout_secs=60)
```
guenp, line 151: Thanks so much for catching this!

guenp (approved):
Thanks @adelebai for fixing these bugs! approved with some suggestions.

azure-quantum/tests/unit/test_job.py
adelebai: I didn't want to do this because it just runs the exact same test as the line above (so slightly redundant)

azure-quantum/tests/unit/test_job.py
adelebai: see comment below - I would prefer to leave it commented out so we don't run the same test twice

PR #123, opened 2021-08-10
Add support for custom TokenFileCredential

xinyi-joffre:
This PR adds support for a custom _TokenFileCredential type that will be used for seamless auth in hosted notebook scenarios.

The credential will only be successful if:
- AZUREQUANTUM_TOKEN_FILE environment variable is set to a valid path
- Token file exists
- Token file contains json with access_token and expires_on in milliseconds
- Token has not expired
- Scope requested is for https://quantum.microsoft.com/.default

In all other cases, the credential will fall back to other methods in the chained credential.

### DIFF
--- a/azure-quantum/azure/quantum/_authentication/_token.py
+++ b/azure-quantum/azure/quantum/_authentication/_token.py
@@ -27,7 +27,7 @@ _TOKEN_FILE_ENV_VARIABLE="AZUREQUANTUM_TOKEN_FILE"
 _AZURE_QUANTUM_SCOPE = "https://quantum.microsoft.com/.default"
 
 
-class _AzureQuantumTokenCredential(object):
+class _TokenFileCredential(object):
     """
     Implements a custom TokenCredential to use a local file as the source for an AzureQuantum token.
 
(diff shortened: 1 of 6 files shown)


anpaz (commented):
My only feedback would be to name the class `TokenFileCredential` instead of `AzureQuantumTokenCredential`.

xinyi-joffre:
> My only feedback would be to name the class `TokenFileCredential` instead of `AzureQuantumTokenCredential`.

Done!

xinyi-joffre:
@anpaz, I removed the scope validation, if you can take another look! Thanks so much for reviewing!

vxfield (approved):
Looks good to me!
Thanks for the misspell fixes too :-)

guenp:
@xinyi-joffre will there be an example or documentation that shows an example of a token credential file or steps on how to generate it?

xinyi-joffre:
> @xinyi-joffre will there be an example or documentation that shows an example of a token credential file or steps on how to generate it?

@guenp, the token credential file should only be generated and used by service in hosted notebook environment, so I don't think we want to have any examples/public documentation around this path. For engineers continuing to maintain/support it, the best place to look is probably the unit tests!

azure-quantum/azure/quantum/_authentication/_token.py
```diff
@@ -0,0 +1,82 @@
...
+            raise CredentialUnavailableError(message="Token file location not set.")
+
+        if not os.path.isfile(self.token_file):
+            raise CredentialUnavailableError(message="Token file at {} does not exist.".format(self.token_file))
+
+        try:
+            token = self._parse_token_file(self.token_file)
+        except JSONDecodeError:
+            raise CredentialUnavailableError(message="Failed to parse token file: Invalid JSON.")
+        except KeyError as e:
```
achocron, line 64: Is there any chance that `str(e)` will expose the token itself? (even if it did, i'm not sure that there'd be any harm in it. and i don't think it would even contain that, but just wanted to double check)

achocron (approved):

PR #124, opened 2021-08-11
Specify encoding on Job submit if submitting compressed blob

anpaz:
Fixes a problem in which the Azure Quantum service will reject jobs if the blob encoding of compressed blobs is not specified.


adelebai (approved):

guenp (approved):
Thanks for catching this!

PR #125, opened 2021-08-12
Workspace.get_targets() returns list of Targets

guenp:
* Workspace.get_targets returns list of available targets
* Add target status info to Target
* Add `Target.refresh()` method to update queue time/availability

Example usage:

```python
[In]
# Get all workspace targets
workspace.get_targets()
```
```stdout
[Out]
[<Target name="ionq.qpu", avg. queue time=5722 s, Available>,
 <Target name="ionq.simulator", avg. queue time=1 s, Available>,
 <Target name="honeywell.hqs-lt-s1", avg. queue time=0 s, Unavailable>,
 <Target name="honeywell.hqs-lt-s1-apival", avg. queue time=1 s, Available>,
 <Target name="honeywell.hqs-lt-s1-sim", avg. queue time=0 s, Available>]
```

```python
[In]
# Get a single target
workspace.get_target("ionq.qpu")
```
```stdout
[Out]
<Target name="ionq.qpu", avg. queue time=5722 s, Available>
```

```
# Refresh the queue time and availability
target.refresh()
```

### DIFF
--- a/azure-quantum/azure/quantum/target/__init__.py
+++ b/azure-quantum/azure/quantum/target/__init__.py
@@ -1,3 +1,8 @@
 from azure.quantum.target.ionq import IonQ
 from azure.quantum.target.honeywell import Honeywell
 from azure.quantum.target.target import Target
+
+ALL_TARGETS = {
+    "ionq": IonQ,
+    "honeywell": Honeywell,
+}
--- a/azure-quantum/azure/quantum/target/honeywell.py
+++ b/azure-quantum/azure/quantum/target/honeywell.py
@@ -16,7 +16,7 @@ class Honeywell(Target):
     def __init__(
         self,
         workspace: Workspace,
-        target: str = "honeywell.hqs-lt-s1-apival",
+        name: str = "honeywell.hqs-lt-s1-apival",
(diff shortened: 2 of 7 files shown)


anpaz (approved):
would it make sense to create a `Target` class instead of using a dict for the values returned by `get_targets`?

guenp:
@anpaz, we already have a `Target` class (for the pass-through). We could indeed adapt that class to include the availability, queue time and status and return a list of instances. e.g.
`workspace.get_targets()` would return `[<IonQ target="ionq.simulator">, <Honeywell target="honeywell.hqs-lt-s1-apival">, ...]`

guenp:
@anpaz I made the changes we discussed. Note that for now, only QC targets are supported; QIO targets are not returned. If we want to support QIO targets or solvers we may want to create a follow up PR. Some options: 1. add a `Workspace.get_solvers()` method 2. create dedicated `Microsoft` target class with hard-coded input and output types 3. add a new API route (if it does not yet exist) that returns the valid input/output data formats and encoding info for a given provider/target and dynamically generate Targets.

guenp:
Closing this in favor of #129

PR #129, opened 2021-08-19
Refactor Solver to inherit from Target, add TargetFactory

### LINKED ISSUE #128: Not all targets have corresponding Target class
In the August release we introduced the `Target` class that is used for submitting raw payloads to Azure Quantum targets. `Target` is a light-weight [abstract base class](https://docs.python.org/3/library/abc.html) that keeps track of the target name, input/output data formats and encoding, and has a convenient `submit` method to submit the payloads and return a `Job` instance.

https://github.com/microsoft/qdk-python/blob/5d424f36e88707d3ecb6277fe0eb0a9f31ff4105/azure-quantum/azure/quantum/target/target.py#L12

Currently, the only providers for which we implemented `Target`s are `IonQ` and `Honeywell`, and we have not implemented `Microsoft` QIO targets. Currently, these targets are implemented using the `Solver` base class.

https://github.com/microsoft/qdk-python/blob/5d424f36e88707d3ecb6277fe0eb0a9f31ff4105/azure-quantum/azure/quantum/optimization/solvers.py#L58

https://github.com/microsoft/qdk-python/pull/125 implements changes to `Workspace.get_targets` to return a list of `Target` objects based on what targets are available in the associated subscription. This currently excludes the QIO targets.

To solve this issue, we suggest to refactor the `Solver` class to inherit from `Target`.

CC @anpaz

guenp:
Fixes #128.
- This PR refactors `Solver` and all of its subclasses to inherit from `Target` and was a consequence of work done in #125.
- All Solvers in `optimization/solvers.py` are now under `azure.quantum.target`; solvers.py is split up into separate files.
- These `Solver` subclasses are now all used by a `TargetFactory` class that is called by `Workspace.get_targets()`.
- Solver.submit() shows a deprecation warning if users passes `compress=False` to avoid the following warning from the service:
```
RuntimeError: Cannot retrieve results as job execution failed(status: Failed.error: {'additional_properties': {}, 'code': 'TransientFailure', 'message': 'The archive entry was compressed using an unsupported compression method.'})
```


azure-quantum/azure/quantum/target/__init__.py
```diff
@@ -1,4 +1,30 @@
-from azure.quantum.target.ionq import IonQ
-from azure.quantum.target.honeywell import Honeywell
-from azure.quantum.target.fleet_management import MicrosoftFleetManagement
-from azure.quantum.target.target import Target
+from .target import Target
```
anpaz, line 1: we should add default targets for other 3pp: toshiba an 1qbit.

azure-quantum/azure/quantum/target/target_factory.py
```diff
@@ -0,0 +1,94 @@
...
+
+class TargetFactory:
+    """Factory class for generating a Target based on a provider and target name"""
+    def __init__(self, workspace: "Workspace"):
+        self.workspace = workspace
+    
+    @staticmethod
+    def _target_cls(provider_id: str, name: str):
+        all_targets = get_all_target_cls()
+
```
anpaz, line 34: this seems expensive, can we cache the results?

azure-quantum/azure/quantum/workspace.py
```diff
@@ -260,18 +265,34 @@ def list_jobs(
...
+        :rtype: Iterable[Target]
         """
-        return {
-            provider.id: [
-                target.id for target in provider.targets
-            ] for provider in self._client.providers.get_status()
-        }
+        from azure.quantum.target.target_factory import TargetFactory
+        target_factory = TargetFactory(workspace=self)
+        return target_factory.get_targets(name=name, provider_id=provider_id)
```
anpaz, line 295: Can we cache the TargetFactory and use it as a singleto?

azure-quantum/azure/quantum/target/target_factory.py
```diff
@@ -0,0 +1,94 @@
...
+        name: _t for t in Target.__subclasses__()
+        for _t in [t] + t.__subclasses__()
+        for name in _t.target_names
+    }
+
+
+class TargetFactory:
+    """Factory class for generating a Target based on a provider and target name"""
+    def __init__(self, workspace: "Workspace"):
+        self.workspace = workspace
```
anpaz, line 29: `Factory` classes tend to be singletons, however, here you create one Factory per Workspace. 
Maybe instead of creating a factory per workspace we push the `workspace` s a parameter of `get_targets`?

azure-quantum/azure/quantum/target/__init__.py
guenp: Thanks for catching this!

azure-quantum/azure/quantum/workspace.py
guenp: Great idea!

azure-quantum/azure/quantum/target/microsoft/fleet_management/__init__.py
```diff
@@ -0,0 +1 @@
+from .fleet_management import FleetManagement
```
anpaz, line 1: Missing Copyright

azure-quantum/azure/quantum/target/microsoft/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .qio import (
```
anpaz, line 1: Missing Copyright

anpaz, line 1: is this needed? my understanding is that this will allow a user to do `from azure.quantum.target.microsoft import ParallelTempering`, but I could be wrong.

anpaz (approved):
left a couple of comments on copyright and imports, but other than that i think this looks great. thanks!

azure-quantum/azure/quantum/target/microsoft/qio/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .parallel_tempering import ParallelTempering
```
anpaz, line 1: missing copyright.

azure-quantum/azure/quantum/target/microsoft/__init__.py
guenp: Yes that's right. I thought it'd be good to be consistent with the other providers, and be able to import `from azure.quantum.target.<provider> import <target>`, the fact that `microsoft` also has a subcategory `qio` and `fleet_management` may not be immediately obvious.

PR #130, opened 2021-08-20
Several fixes in the test recording processor

vxfield:
This PR fixes a few issues with the unit test Python VCR recording/playback processors that are used to sanitize the recordings:

1) Reuse common Azure SDK environment variables as an alternative to the Azure Quantum environment variables for automated testing
 ```python
        self._resource_group = os.environ.get("AZUREQUANTUM_WORKSPACE_RG", os.environ.get("RESOURCE_GROUP", RESOURCE_GROUP))
        self._subscription_id = os.environ.get("AZUREQUANTUM_SUBSCRIPTION_ID", os.environ.get("SUBSCRIPTION_ID", ZERO_UID))
        self._workspace_name = os.environ.get("AZUREQUANTUM_WORKSPACE_NAME")
        self._location = os.environ.get("AZUREQUANTUM_WORKSPACE_LOCATION", os.environ.get("LOCATION", LOCATION))
```

2) More aggressive Blob Storage Shared Access Key sanitization (the previous regex was missing some keys that contained `/`):
```python
regex_replacer.register_regex(r"sig=[^&]+\&", "sig=PLACEHOLDER&")
```
 
3) Fixture and helper method for upcoming async functionality
```python
@pytest.mark.usefixtures("event_loop_instance")

def get_async_result(self, coro):
        return asyncio.get_event_loop().run_until_complete(coro)
```

4) Removal of `content-disposition` in the header allow-list as it leaks file names in the recording and it's not needed.

5) Improved logic to handle body regex replacements that was causing some exceptions and failing the EnvironmentCredential authentication during recordings: 
```python
body = request.body
        if body is not None:
            if ((content_type == "application/x-www-form-urlencoded") and
                (isinstance(body, bytes) or isinstance(body, bytearray))):
                body = body.decode("utf-8")
                body = self.regex_replace_all(body)
                request.body = body.encode("utf-8")
            else:
                body = str(body)
                body = self.regex_replace_all(body)
                request.body = body
```

6) Sanitization of response `url` field that can show up in some responses:
```python
if "url" in response:
            response["url"] = self.regex_replace_all(response["url"])
```


anpaz (approved):
lgtm

PR #131, opened 2021-08-20
Changing schema of grouped terms and separating term types

adelebai:
The solutions team recently decided on a schema change on how grouped terms are supplied to the backend. 
For this reason we need to update the SDK side. This needs to make it by the August shiproom.  

Changing schema of grouped terms and separating term types. 
The new schema will also no longer work easily with streamingproblem.py so I am removing support there for now.

### DIFF
--- a/azure-quantum/azure/quantum/optimization/problem.py
+++ b/azure-quantum/azure/quantum/optimization/problem.py
@@ -13,7 +13,7 @@ import os
 
 from typing import List, Tuple, Union, Dict, Optional, Type, TYPE_CHECKING
 from enum import Enum
-from azure.quantum.optimization import TermBase, Term, GroupType, GroupedTerm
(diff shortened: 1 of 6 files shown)


vxfield (changes requested):
Looks good overall but I think you need to run the unit tests and update the recordings for this scenario.
Please check the readme in the `tests` folder and let me know if you have questions.

adelebai:
@Xfield I reuploaded the recordings files for the solvers.

vxfield:
Hi @adelebai ,
It looks like some of the recordings were not made from a successful run.
Before checking-in the recordings, make sure that the unit tests pass when doing the recordings, otherwise they will also fail in the CI build that runs them in playback mode.
See the Readme in the `tests` folder for instructions on setting environment variables to specify the workspace and credentials and ping me if you have questions or find any challenge.

PR #132, opened 2021-08-24
Add deprecation notice to 3rd party QIO namespaces

guenp:


azure-quantum/azure/quantum/optimization/oneqbit/solvers.py
```diff
@@ -0,0 +1,14 @@
+# coding=utf-8
+##
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+##
+import warnings
+warnings.warn("The azure.quantum.optimization.oneqbit namespace will be deprecated. \
```
xinyi-joffre: It looks like with latest qdk build, we support both of the below:

Option 1: from azure.quantum.target.oneqbit import PathRelinkingSolver
Option 2: from azure.quantum.optimization import PathRelinkingSolver

Do we prefer option 1? In which case, should we remove the option 2 path, which is new for the August release?

guenp: @xinyi-joffre, I agree, we should deprecate option 2, this was added in error. Thanks for pointing this out!

Issue #118, opened 2021-08-03
Support Unit Test Recording playback for multiple calls


guenp:
Update `QuantumTestBase` to support repeated calls in unit test playback mode.

See note here:

https://github.com/microsoft/qdk-python/blob/f7e944d6f674e5aa63f63f8f0d3c84ede2a73502/azure-quantum/tests/unit/test_job.py#L245

guenp:
A workaround for this issue is to mock out the `Job.wait_until_completed` method with a mock wait method that pauses/resumes the recording before and after:

```python
with mock.patch.object(
    Job,
    "wait_until_completed",
    self.mock_wait(Job.wait_until_completed)
):
    service.run(...)
```

with a method `mock_wait` on the `QuantumTestBase` class:

```python
    def mock_wait(self, job_wait_until_completed):
        # Workaround for issue #118
        # See: https://github.com/microsoft/qdk-python/issues/118
        # Create closure to expose self and Job.wait_until_completed to
        # mock function
        def _mock_wait(job, *args, **kwargs):
            # Pause recording such that we don't record multiple calls
            self.pause_recording() 
            job_wait_until_completed(job, *args, **kwargs)
            self.resume_recording()

            # Record a single GET request such that job.wait_until_completed
            # doesn't fail when running recorded tests
            job.refresh()

        return _mock_wait
```

PR #145, opened 2021-09-13
[WIP] Get Workspace from name only

guenp:
Add a way to retrieve a `Workspace` from only the name without requiring the resource ID, subscription ID, resource group or location.
Note that to add a UT that tests passing no subscription ID requires fixing #118.


azure-quantum/azure/quantum/workspace.py
```diff
@@ -189,6 +206,93 @@ def __init__(
         # Create QuantumClient
         self._client = self._create_client()
 
+    @classmethod
+    def from_name(
```
guenp, line 210: Make private

guenp:
TODO: add in workaround for TokenFileCredential failure

guenp:
Closing this in favor of a future endpoint URI implementation.

PR #149, opened 2021-09-22
Add Qiskit and Cirq plugins for IonQ and Honeywell targets

guenp:
This feature branch adds support for submitting Qiskit and Cirq circuits to IonQ and Honeywell.

Example usage for Qiskit:

```python
from qiskit import QuantumCircuit
from azure.quantum.qiskit import AzureQuantumProvider

# Enter your Azure Quantum workspace details here
provider = AzureQuantumProvider(resource_id="", location="")

backend = provider.get_backend("ionq.qpu")
job = backend.run(your_qiskit_circuit, shots=1000)
result = job.result()
```

Example usage for Cirq:

```python
import cirq
import azure.quantum.cirq as aq

# Enter your Azure Quantum workspace details here
service = aq.AzureQuantumService(resource_id="", location="")
result = service.run(your_cirq_circuit, repetitions=1000, target="ionq.qpu")
```


azure-quantum/azure/quantum/plugins/cirq/targets/ionq.py
```diff
@@ -0,0 +1,170 @@
...
+        param_resolver: cirq.ParamResolverOrSimilarType = cirq.ParamResolver({}),
+        seed: cirq.RANDOM_STATE_OR_SEED_LIKE = None,
+    ) -> "cirq.Result":
+        if isinstance(result, QPUResult):
+            return result.to_cirq_result(params=cirq.ParamResolver(param_resolver))
+        elif isinstance(result, SimulatorResult):
+            return result.to_cirq_result(params=cirq.ParamResolver(param_resolver), seed=seed)
+
+        raise ValueError("Result {result} not supported. \
+            Please pass either a cirq_ionq.results.QPUResult \
```
anpaz, line 168:
```suggestion
            Expecting either a cirq_ionq.results.QPUResult \
```
The user is not calling this method directly, so asking them to pass certain values is missleading.

azure-quantum/azure/quantum/plugins/cirq/service.py
```diff
@@ -0,0 +1,201 @@
...
+
+        return target_factory
+
+    def targets(
+        self,
+        name: str = None,
+        provider_id: str = None,
+        **kwargs
+    ) -> Union["CirqTarget", List["CirqTarget"]]:
+        """Get all targets available in the Azure Quantum.
```
anpaz, line 71:
```suggestion
        """Get all quantum computing targets available in the Azure Quantum Workspace.
```

azure-quantum/azure/quantum/plugins/cirq/service.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def create_job(
+        self,
+        program: cirq.Circuit,
+        repetitions: int,
+        name: str = DEFAULT_JOB_NAME,
+        target: str = None,
+        param_resolver: cirq.ParamResolverOrSimilarType = cirq.ParamResolver({})
+    ) -> Union["CirqJob", "CirqIonqJob"]:
+        """Create job using a Cirq program
```
anpaz, line 112:
```suggestion
        """Create job to run the given `cirq` program in Azure Quantum
```

azure-quantum/azure/quantum/plugins/cirq/service.py
```diff
@@ -0,0 +1,201 @@
...
+        self,
+        program: cirq.Circuit,
+        repetitions: int,
+        target: str = None,
+        name: str = DEFAULT_JOB_NAME,
+        param_resolver: cirq.ParamResolverOrSimilarType = cirq.ParamResolver({}),
+        seed: cirq.RANDOM_STATE_OR_SEED_LIKE = None,
+        timeout_seconds: int = DEFAULT_TIMEOUT,
+    ) -> cirq.Result:
+        """Run Cirq circuit on specified target, if target not specified run on default target
```
anpaz, line 153:
```suggestion
        """Run Cirq circuit on specified target, if target not specified then it runs on the default target
```

azure-quantum/azure/quantum/plugins/cirq/targets/honeywell.py
```diff
@@ -0,0 +1,102 @@
...
+            output_data_format=output_data_format,
+            provider_id=provider_id,
+            content_type=content_type,
+            encoding=encoding,
+            **kwargs
+        )
+
+    @staticmethod
+    def _translate_cirq_circuit(circuit) -> str:
+        """Translate Cirq circuit to Honeywell QASM."""
```
anpaz, line 46:
```suggestion
        """Translate `cirq` circuit to QASM."""
```

azure-quantum/azure/quantum/plugins/cirq/targets/ionq.py
```diff
@@ -0,0 +1,170 @@
...
+            content_type=content_type,
+            encoding=encoding,
+            **kwargs
+        )
+
+    @staticmethod
+    def _translate_cirq_circuit(circuit) -> Dict[str, Any]:
+        """Translate Cirq circuit to IonQ JSON. If dependencies \
+    are not installed, throw error with installation instructions."""
+        try:
```
anpaz, line 106: is the `try` necessary, we already that at the top of the module

azure-quantum/azure/quantum/plugins/cirq/targets/ionq.py
```diff
@@ -0,0 +1,170 @@
...
+
+class _IonQClient:
+    """Thin wrapper around Workspace to support cirq_ionq.Job"""
+    def __init__(self, workspace: "Workspace"):
+        self._workspace = workspace
+    
+    @staticmethod
+    def _to_ionq_status(status: str):
+        from azure.quantum._client.models._quantum_client_enums import JobStatus
+        _STATUS_DICT = {
```
anpaz, line 32: isn't this more broadly needed? i'm particular concerned if we need to add a new status, we shouldn't need to update more than one file.

azure-quantum/azure/quantum/plugins/qiskit/backends/ionq.py
```diff
@@ -0,0 +1,164 @@
...
+    def _job_metadata(self, circuit, meas_map):
+        return {
+            "qiskit": True,
+            "name": circuit.name,
+            "num_qubits": circuit.num_qubits,
+            "meas_map": meas_map,
+        }
+
+    def run(self, circuit, **kwargs):
+        """Submits the given circuit for execution on an IonQ target."""
```
anpaz, line 49:
```suggestion
        """Submits the given circuit to run on an IonQ target."""        
```

azure-quantum/tests/unit/test_plugins.py
```diff
@@ -0,0 +1,294 @@
...
+            # playback currently does not work for repeated calls
+            # See: https://github.com/microsoft/qdk-python/issues/118
+            if not self.is_playback:
+                job = qiskit_job._azure_job
+                self.assertEqual(False, job.has_completed())
+                if self.in_recording:
+                    import time
+                    time.sleep(3)
+                job.refresh()
+                job.wait_until_completed()
```
anpaz, line 73: if we're calling this, why do we need the time.sleep(3) above?

anpaz: we should pass a timeout here.

azure-quantum/tests/unit/test_plugins.py
```diff
@@ -0,0 +1,294 @@
...
+                self.assertEqual(True, job.has_completed())
```
anpaz, line 74: For live tests, it is not safe to check if a job finishes, as it can timeout when running on 3PP. Better not fail the test if job is not completed.

azure-quantum/tests/unit/test_plugins.py
```diff
@@ -0,0 +1,294 @@
...
+                job = provider.get_job(job.id)._azure_job
+                self.assertEqual(True, job.has_completed())
+
+            result = qiskit_job.result()
```
anpaz, line 78: should only call `result()` if job has completed.

azure-quantum/tests/unit/test_plugins.py
```diff
@@ -0,0 +1,294 @@
...
+                circuit=circuit,
+                shots=num_shots
+            )
+
+            # Make sure the job is completed before fetching the results
+            # playback currently does not work for repeated calls
+            # See: https://github.com/microsoft/qdk-python/issues/118
+            if not self.is_playback:
+                job = qiskit_job._azure_job
+                self.assertEqual(False, job.has_completed())
```
anpaz, line 68: we should be testing the corresponding methods on the qiskit_job, not on the azure job.

azure-quantum/tests/unit/test_plugins.py
```diff
@@ -0,0 +1,294 @@
...
+            qiskit_job = backend.run(
+                circuit=circuit,
+                num_shots=100
+            )
+            job = qiskit_job._azure_job
+
+            # Make sure the job is completed before fetching the job
+            if not self.is_playback:
+                self.assertEqual(False, job.has_completed())
+                if self.in_recording:
```
anpaz, line 105: same comments as above:
- should be using qiskit_job and not the azure job's methods
- should not expect job will actually complete
- should add a timeout.
more over, for this test, is not even necessary wait for job to complete as all we're doing is making sure we can retrieve an existing job.

azure-quantum/azure/quantum/plugins/cirq/targets/honeywell.py
```diff
@@ -0,0 +1,102 @@
...
+            for key, value in result.items()
+            if key.startswith("m_")
+        }
+        return Result(params=param_resolver, measurements=measurements)
+    
+    def _to_cirq_job(self, azure_job: "AzureJob", program: "cirq.Circuit" = None):
+        """Convert Azure job to Cirq job"""
+        if program is None:
+            # Download QASM and convert to Cirq circuit
+            from cirq.contrib.qasm_import import circuit_from_qasm
```
anpaz, line 63: why is this only necessary for Honeywell and not IonQ?

azure-quantum/tests/unit/test_plugins.py
```diff
@@ -0,0 +1,294 @@
...
+                num_shots=num_shots
+            )
+
+            job = qiskit_job._azure_job
+
+            # Make sure the job is completed before fetching the results
+            # playback currently does not work for repeated calls
+            # See: https://github.com/microsoft/qdk-python/issues/118
+            if not self.is_playback:
+                self.assertEqual(False, job.has_completed())
```
anpaz, line 144: same comments as above:
- should be using qiskit_job and not the azure job's methods
- should not expect job will actually complete
- should add a timeout.

azure-quantum/tests/unit/test_plugins.py
```diff
@@ -0,0 +1,294 @@
...
+                if self.in_recording:
+                    import time
+                    time.sleep(3)
+                job.refresh()
+                job.wait_until_completed()
+                self.assertEqual(True, job.has_completed())
+                job = workspace.get_job(job.id)
+                self.assertEqual(True, job.has_completed())
+
+            result = qiskit_job.result()
```
anpaz, line 154: should only check results if job completed.

anpaz (changes requested):

azure-quantum/azure/quantum/plugins/cirq/service.py
```diff
@@ -0,0 +1,201 @@
...
+        :param param_resolver: Cirq parameters, defaults to cirq.ParamResolver({})
+        :type param_resolver: cirq.ParamResolverOrSimilarType, optional
+        :param seed: Random seed for simulator results, defaults to None
+        :type seed: cirq.RANDOM_STATE_OR_SEED_LIKE, optional
+        :param timeout_seconds: Timeout in seconds, defaults to None
+        :type timeout_seconds: int, optional
+        :return: Measurement results
+        :rtype: cirq.Result
+        """
+        # Get target
```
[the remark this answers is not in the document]
guenp: It is, but we still need it here to translate the results to a `cirq.Result` object.

azure-quantum/azure/quantum/plugins/cirq/service.py
```diff
@@ -0,0 +1,201 @@
...
+            name=name,
+            target=target,
+            param_resolver=param_resolver
+        )
+        # Get raw job results
+        try:
+            result = job.results(timeout_seconds=timeout_seconds)
+        except RuntimeError as e:
+            # Catch errors from cirq_ionq.Job.results
+            if "Job was not completed successful. Instead had status: " in str(e):
```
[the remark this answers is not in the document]
guenp: This was causing confusion.

azure-quantum/azure/quantum/plugins/cirq/service.py
```diff
@@ -0,0 +1,201 @@
...
+                raise TimeoutError(f"The wait time has exceeded {timeout_seconds} seconds. \
```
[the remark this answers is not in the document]
guenp: If the job failed, it will just raise on line 194.

anpaz (approved):

azure-quantum/azure/quantum/plugins/cirq/targets/honeywell.py
guenp: The reason we need to pull this data from the service and translate back to Cirq is because the IonQ translation provides metadata that is stored in the job metadata. We can do the same for Honeywell.

azure-quantum/azure/quantum/plugins/cirq/targets/ionq.py
guenp: It is only needed here because it is specific to the way the cirq_ionq.Job object handles the job states. This mapping currently happens on the service side so here we are simply mapping it back to the original IonQ status values.

azure-quantum/azure/quantum/plugins/cirq/targets/ionq.py
guenp: Yes I agree it's redundant and can be removed. Thanks for catching that.

azure-quantum/tests/unit/test_plugins.py
guenp: Here we can actually use the new pause_recording feature.

azure-quantum/tests/unit/test_plugins.py
guenp: I think we do need to wait for it to complete, we are looking if we can get an existing job that has completed. We could just submit the job and retrieve it but then we can't check if `job.has_completed()` is the same for both jobs as it could have changed between calls.

PR #154, opened 2021-09-30
Update README, add Azure Quantum logo

guenp:
Update the README to have more examples and links to the docs.
Azure Quantum logo in header for added flair.


guenp:
Note that to be able to see the image on PyPI, we need to update the link to the image to include an absolute link to the file on github. I will create a follow up PR for this after the image has been merged to `main`.

azure-quantum/README.md
```diff
@@ -1,18 +1,108 @@
-# Python SDK for Azure Quantum #
+![Azure Quantum logo](Azure-Quantum-logo.png)
```
cgranade, line 1: Will this logo need to be included in the package for PyPI's Markdown renderer to pick it up?

guenp: Yes, see comment above. We also need to add a direct link to the file. Thanks for your review!

cgranade: Got it, I hadn't seen that when I posted, sorry.

anpaz (commented):
I think the documentation looks good, however, I would prefer to keep the development instructions in the README.md, as that's where a developer would normally look for when browing the repository, and move the package description to a `DESCRIPTION.md` file.

guenp:
> I think the documentation looks good, however, I would prefer to keep the development instructions in the README.md, as that's where a developer would normally look for when browing the repository, and move the package description to a `DESCRIPTION.md` file.

This is a good point; the development instructions should be more visible. However, I'm not sure we need to include all of that content in the README. I suggest to use a CONTRIBUTING.md file instead and link to that from the README.

anpaz (approved):

PR #159, opened 2021-10-14
Add protobuf support to qdk

sanjgupt:
**NOTE: Introducing a breaking change**: In problem class' deserialize method, the parameter problem_as_json of type string
is changed to input_problem of type Union[str, list] to accomodate both json strings and protobuf messages.

This feature is to support protobuf serialization of problems. 
When a user specifies a problem, they can now create Problem(name=<>, problem_type=<>, serialization_type="application/x-protobuf") which is an optional value. If not supplied, json is assumed and no change to the current flow occurs except the following:
1.  when calling problem.serialize the serialize function calls serialize_to_json and 
2. problem.deserialize calls deserialize_from_json
3. When uploading the problem, blob's metadata is set to "{SerializtionType:application/json}"

If set to application/x-protobuf the following things occur: 
1. problem.serialize calls serialize_to_protobuf. This method, constructs the problem as a list of protobuf messages, each message containing 1000 terms and returns a list. The name field in the metadata is added in this step. The problem type and the name are attached to the first message in the list.
2. problem.deserilize forwards to deserialize_from_json: This method reads the list of messages and constructs the Problem object
3. When uploading, the method sets the blob's metadata property to {"SerializationType": application/x-protobuf} 
4. When downloading, the method checks for the SerializationType metadata set on the blob and if it's set to application/x-protobuf, it calls the  deserialize_from_proto method to complete it's operation
5. When submitting a job with Solver.submit(), there is a check for if the specified solver supports protobuf. If not, a value error is returned to the user

The following solvers are supported:
1. SubstochasticMonteCarlo
2. PopulationAnnealing
Both these solver classes now have protobuff_supported method that returns true. 

In addition, a problem.proto file is added, which is the schema for the format and a problem_pb2.py which is generated from this schema.

The following test scenarios are added: 
1. serializing to proto
2. deserialing from proto
3. downloading a proto
4. Submitting a solver with valid solver type for proto
5. Submitting a solver with invalid sovler type for proto and get user error.

The following diagram explains how the protobuf messages are sent from the problem as a series of smaller messages with 
1000 terms and the serialize_to_proto methd in problem.py does this construction

image


azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -211,6 +215,8 @@ def upload_input_data(
         content_type: str,
         blob_name: str = "inputData",
         encoding: str = "",
+        #serialization_type: Optional[str] = "application/json",
+        serialization_type: str = "application/json",
```
anpaz, line 219: again, this should be saved in the `content_type` property of the blob.

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -59,12 +70,14 @@ def __init__(
         terms: Optional[List[TermBase]] = None,
         init_config: Optional[Dict[str, int]] = None,
         problem_type: ProblemType = ProblemType.ising,
+        serialization_type: Optional[str] = None
```
anpaz, line 73: I feel this one should be an enum (like `ProblemType`) as opposed to a free-form string, as the code will take specific actions based on its value.

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -98,34 +120,59 @@ def serialize(self) -> str:
...
+                    for j in range (len(self.terms[i].ids)):
+                        term.ids.append(self.terms[i].ids[j])
+            msg_count += 1
+            terms_remaining -= 1000
+            proto_messages.append(proto_problem.SerializeToString())
+        return proto_messages
+    
     @classmethod
-    def deserialize(
+    def deserialize_from_json(
```
anpaz, line 170: i think the convention is to make this a static method and just call it `from_json`

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -147,6 +194,67 @@ def deserialize(
             problem.init_config = result["cost_function"]["initial_configuration"]
 
         return problem
+    
+    @classmethod
+    def deserialize_from_proto(
```
anpaz, line 199: `from_proto`

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -147,6 +194,67 @@ def deserialize(
...
+        problem_str: Union[list,str], 
+        name:Optional[str] = None, 
+        serialization_type:Optional[str] = None) -> Problem:
+        """Deserializes the problem from a
+        JSON string or protobuf messages serialized with Problem.serialize()
+        Also used to deserialize the messages downloaded from the blob
+
+        :param problem_str:
+            The string to be deserialized to a `Problem` instance
+        :type problem_str: str
```
anpaz, line 244: is the type for a problem serialized as protobuf really `str`? I was expecting this to be just Byte[]

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -98,34 +120,59 @@ def serialize(self) -> str:
...
+            else:
+                # add the remaining terms to the last message
+                for i in range(terms_remaining):
+                    term = cost_function.terms.add()
+                    term.c = self.terms[i].c
+                    for j in range (len(self.terms[i].ids)):
+                        term.ids.append(self.terms[i].ids[j])
+            msg_count += 1
+            terms_remaining -= 1000
+            proto_messages.append(proto_problem.SerializeToString())
```
anpaz, line 166: Is `ToString` the most efficient way to do this? Since the goal is to make this efficient, shouldn't we just use the raw bytes?

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -147,6 +194,67 @@ def deserialize(
             problem.init_config = result["cost_function"]["initial_configuration"]
 
         return problem
+    
+    @classmethod
+    def deserialize_from_proto(
+        cls,
+        problem_as_str: Union[list,str],
```
anpaz, line 201: this is just a `list`, isn't it?

anpaz: the name of the variable is missleading as it's not a string.

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -222,19 +330,36 @@ def to_blob(self, compress: bool = False) -> bytes:
...
         """
-        problem_json = self.serialize()
-        logger.debug("Problem json: " + problem_json)
+        input_problem = self.serialize()
+        debug_input_string = input_problem if type(input_problem) is str else b''.join( input_problem).decode()
+        logger.debug("Input Problem: " + debug_input_string)
         data = io.BytesIO()
 
-        if compress:
+        if self.serialization_type == "application/x-protobuf":
```
anpaz, line 342: this should go into the existing `serialize_to_proto` as it only applies to protobuf

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -274,6 +399,9 @@ def upload(
             blob_name = self._blob_name()
 
         encoding = "gzip" if compress else ""
+        content_type = "application/json"
```
anpaz, line 402: use only `content_type` to specify serialization.

azure-quantum/azure/quantum/problem.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
```
anpaz, line 1: this probably should go in the `quantum/optimization` folder too

azure-quantum/azure/quantum/storage.py
```diff
@@ -102,15 +103,21 @@ def upload_blob(
     content_settings = ContentSettings(
         content_type=content_type, content_encoding=content_encoding
     )
+    
+    metadata = {
```
anpaz, line 107: use only `content_type`.

azure-quantum/azure/quantum/target/solvers.py
```diff
@@ -238,13 +242,27 @@ def check_valid_problem(self, problem):
                     f"with problem type {problem.problem_type};"
                     f"Try PopulationAnnealing or SubstochasticMonteCarlo."
                 )
+        if problem.serialization_type == "application/x-protobuf":
+            if not self.supports_protobuf() and self.name not in proto_valid_solver_names:
+                raise ValueError(
+                    f"Solver type is not compatible"
```
anpaz, line 248: Please add
```suggestion
                    f"Solver `{self.name} type is not compatible "
```

azure-quantum/azure/quantum/target/solvers.py
```diff
@@ -238,13 +242,27 @@ def check_valid_problem(self, problem):
...
+                    f"for serialization with protobuf;"
```
anpaz, line 249:
```suggestion
                    f"for serialization with protobuf; "
```

azure-quantum/azure/quantum/target/solvers.py
```diff
@@ -238,13 +242,27 @@ def check_valid_problem(self, problem):
...
+                    f"Try PopulationAnnealing or SubstochasticMonteCarlo."
```
anpaz, line 250: Is good to point the user to this, however the problem is that it might change and we will forget to update it :(

azure-quantum/environment.yml
```diff
@@ -17,9 +17,10 @@ dependencies:
   - aiohttp>=3.7.0
   - pip:
     - azure-devtools>=1.2.0
-    - cirq>=0.12.0
-    - cirq-ionq>=0.12.0
+    #- cirq>=0.12.0
```
anpaz, line 20: i'm pretty sure we still need these.

azure-quantum/tests/unit/test_problem.py
```diff
@@ -107,32 +110,32 @@ def test_problem_name_serialization(self):
             assert problem_json["metadata"]["name"] == problem_name
 
             # deserializes name
-            deserialized_problem = Problem.deserialize(problem_as_json=serialized_problem)
+            deserialized_problem = Problem.deserialize(problem_str=serialized_problem)
```
anpaz, line 113: the fact that you're changing the parameter name is a breaking change. You should add a new parameter and keep the old one.

azure-quantum/tests/unit/test_problem.py
```diff
@@ -346,8 +361,62 @@ def test_terms_from_npz_pubo(self):
...
+            Term(c=5, indices=[2, 0]),
+            Term(c=3, indices=[1, 0]),
+            Term(c=5, indices=[2, 0]),
+        ]
+        problem_msgs = problem.serialize()
+        deserialized_problem = Problem.deserialize(problem_msgs)
+        self.assertEqual( len(deserialized_problem.terms), 12 )
+        self.assertEqual(deserialized_problem.problem_type, ProblemType.pubo)
+        self.assertEqual(deserialized_problem.name, problem.name)
+        #self.assertEqual( len(deserialized_problem.terms), 12 )
```
anpaz, line 418: Can we add a test that checks that if we upload a problem with the `application/x-protobuf` encoding, it can be deserialized without having to pass the encoding itself?

azure-quantum/tests/unit/test_solvers.py
```diff
@@ -44,3 +44,24 @@ def test_number_of_solutions_set(testsolver):
     testsolver.set_number_of_solutions(100)
     assert param_name in testsolver.params["params"]
     assert testsolver.params["params"][param_name] == 100
+
+def test_submit_proto_problem(self):
```
anpaz, line 48: is this a live test? we should add the @pytest.live mark and probably the @pytest.qio mark.
If it needs a special region and/or provider the capability that controls the test might be different.
If this is not a live test, we should add a couple.

anpaz (changes requested):

azure-quantum/azure/quantum/__init__.py
```diff
@@ -10,5 +10,7 @@
 
 from ._client.models._quantum_client_enums import JobStatus
 
+from .problem_pb2 import *
```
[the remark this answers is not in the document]
guenp: I agree, some explicit imports would be helpful here to not introduce unintended breaking changes in the future.

azure-quantum/azure/quantum/problem_pb2.py
```diff
@@ -0,0 +1,277 @@
+# -*- coding: utf-8 -*-
```
guenp, line 1: I suggest to move this file to a subpackage, e.g. azure/quantum/serialization/protobuf.py

azure-quantum/azure/quantum/target/microsoft/qio/population_annealing.py
```diff
@@ -77,4 +77,7 @@ def __init__(
         self.check_set_positive_int("timeout", timeout)
 
     def supports_grouped_terms(self):
+        return True
+    
+    def supports_protobuf(self):
```
guenp, line 82: Please add this method to the `Target` class to return `False` by default.

azure-quantum/azure/quantum/target/solvers.py
```diff
@@ -238,13 +242,27 @@ def check_valid_problem(self, problem):
...
+                )
 
     def supports_grouped_terms(self):
         """
         Return whether or not the Solver class supported grouped terms in the cost function.
         This should be overridden by Solver subclasses which do support grouped terms.
         """
         return False
+    
+    def supports_protobuf(self):
```
guenp, line 260: I would add this to `Target` instead of `Solver`

guenp (changes requested):
Left some comments.
Please rename `azure.quantum.problem_pb2` to something else, like `azure.quantum.serialization.protobuf`.

azure-quantum/azure/quantum/target/solvers.py
sanjgupt: I'll add a running task for myself to track these :)

azure-quantum/tests/unit/test_problem.py
sanjgupt: Oh right, got it! I'll update

azure-quantum/tests/unit/test_problem.py
sanjgupt: Hi, if the content type is not specified when creating problem, a json problem will be created. 
In this test, the content type protobuf is specified only for creating the problem, the deserialzition function is called without specifying the content type. Is that what you were expecting to test?

sanjgupt:
> Left some comments. Please rename `azure.quantum.problem_pb2` to something else, like `azure.quantum.serialization.protobuf`.

Hi, I moved the file to serialization, but not renaming it since the name is generated based on the protobuf format schema

azure-quantum/azure/quantum/problem.proto
sanjgupt: I moved it to the serialzition folder with the serialzition code

azure-quantum/azure/quantum/optimization/problem.py
sanjgupt: This a protobuf generated function which is creating byte strings. This is an unfortunate name but it's gen code

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -208,7 +208,7 @@ def from_storage_uri(
     def upload_input_data(
         container_uri: str,
         input_data: bytes,
-        content_type: str,
+        content_type: str = "application/json",
```
anpaz, line 211: i'm not sure if this should be an optional parameter. only the caller knows the right content_type on the blob.

azure-quantum/azure/quantum/serialization/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .problem_pb2 import (
```
anpaz, line 1: Missing Copyright header.

azure-quantum/azure/quantum/serialization/problem.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
```
anpaz, line 1: Can we add comments? If so we should add copyright and other comments to document the datastructure (or pointer to the documentation)

azure-quantum/environment.yml
```diff
@@ -23,3 +23,4 @@ dependencies:
     - qiskit-terra>=0.18.0
     - asyncmock>=0.4.0
     - aiofile>=3.7.2
+    - protobuf==3.14.0
```
anpaz, line 26: wonder if this should be an optional dependency

anpaz (approved):
Left a couple of minor comments, otherwise lgtm.

azure-quantum/azure/quantum/serialization/__init__.py
sanjgupt: This is generated code. 
We should not edit it

azure-quantum/azure/quantum/serialization/problem.proto
sanjgupt: Yes, this file is editable. Adding comments and copyright information

azure-quantum/azure/quantum/job/base_job.py
guenp: I agree, also I suggest to change the type from `str` to `ContentType`

azure-quantum/environment.yml
sanjgupt: Hmm, I have added some code that takes dependency on google protobuf functions. That code breaks without having this available.

azure-quantum/azure/quantum/job/base_job.py
sanjgupt: This creates a circular dependency b/w base_job and problem code files. Is there another way to do it without importing the ContentType from problem?

guenp (changes requested):

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -28,13 +35,20 @@
...
     ising = 1
     pubo_grouped = 2
     ising_grouped = 3
 
+proto_types = {
+    ProblemType.ising: ProtoProblem.ProblemType.ISING,
+    ProblemType.pubo: ProtoProblem.ProblemType.PUBO, 
+}
+
+class ContentType(str, Enum):
```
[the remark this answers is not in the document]
sanjgupt: That sounds good! Updating

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -219,8 +223,8 @@ def upload_input_data(
         :type container_uri: str
         :param input_data: Input data in binary format
         :type input_data: bytes
-        :param content_type: Content type, e.g. "application/json"
-        :type content_type: str
+        :param content_type: Content type, e.g. "application/json" or application/x-protobuf
```
guenp, line 226:
```suggestion
        :param content_type: Content type, e.g. "application/json" or "application/x-protobuf"
```

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -98,35 +123,79 @@ def serialize(self) -> str:
...
+            msg_count += 1
+            terms_remaining -= 1000
+            proto_messages.append(proto_problem.SerializeToString())
+        return proto_messages
+    
+    def compress_protobuf(
+        self, 
+    proto_messages: List[str] ) -> bytes:
+    # Write to a series of files to folder and compress
+    # QIOTE expects all files to be names "gzipinputfile_pb_{file_count}.pb" at this time
```
guenp, line 176:
```suggestion
    def compress_protobuf(
        self, 
        proto_messages: List[str] ) -> bytes:
    """
    Write to a series of files and compress the files in a tar file.
    
    :param proto_messsages: A list of file names, currently should be in the format "gzipinputfile_pb_{file_count}.pb"
    """
```

guenp (changes requested):

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -147,6 +216,68 @@ def deserialize(
...
+        cls,
+        problem_msgs: list,
+        name: Optional[str] = None
+    ) -> Problem:
+        msg_count = 0
+
+        problem = cls(
+            name = name
+        )
+
```
guenp, line 231: Missing docstring here

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -147,6 +216,68 @@ def deserialize(
...
+                    term.ids.append(msg_term_id)
+                problem.terms.append(term)
+        
+        return problem
+
+    @classmethod
+    def deserialize(
+        cls, 
+        input_problem: Union[str, list],
+        name:Optional[str] = None,
```
guenp, line 258:
```suggestion
        name: Optional[str] = None, 
```

azure-quantum/azure/quantum/optimization/problem.py
```diff
@@ -147,6 +216,68 @@ def deserialize(
...
+                problem.terms.append(term)
+        
+        return problem
+
+    @classmethod
+    def deserialize(
+        cls, 
+        input_problem: Union[str, list],
+        name:Optional[str] = None, 
+        content_type:Optional[ContentType] = None) -> Problem:
```
guenp, line 259:
```suggestion
        content_type: Optional[ContentType] = None) -> Problem:
```

guenp (approved):

PR #180, opened 2021-11-05
fail test.ps1 script if pytest did not return zero exit code

guenp:
This fixes the issue that failing unit tests are not causing build steps to fail. See: https://ms-quantum.visualstudio.com/Quantum%20Program/_build/results?buildId=174305&view=logs&j=0cf5fac5-e08e-5c92-a2db-66aab767ad9a&t=3561fc1c-61af-5606-6b74-430059fea15f&l=131

### DIFF
--- a/build/test.ps1
+++ b/build/test.ps1
@@ -24,5 +24,8 @@ if ($Env:ENABLE_PYTHON -eq "false") {
   foreach ($PackageName in $PackageNames) {
     $EnvName = GetEnvName -PackageName $PackageName -CondaEnvironmentSuffix $CondaEnvironmentSuffix
     Invoke-Tests -PackageName $PackageName -EnvName $EnvName
+    if (0 -ne $LastExitCode) {
+      exit 1
+    }
   }
 }
--- a/build/test.ps1
+++ b/build/test.ps1
@@ -24,10 +24,12 @@ if ($Env:ENABLE_PYTHON -eq "false") {
   $ExitCode = 0;
   foreach ($PackageName in $PackageNames) {
     $EnvName = GetEnvName -PackageName $PackageName -CondaEnvironmentSuffix $CondaEnvironmentSuffix
-    $Success = Invoke-Tests -PackageName $PackageName -EnvName $EnvName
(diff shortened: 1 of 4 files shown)


build/test.ps1
```diff
@@ -24,5 +24,8 @@ if ($Env:ENABLE_PYTHON -eq "false") {
   foreach ($PackageName in $PackageNames) {
     $EnvName = GetEnvName -PackageName $PackageName -CondaEnvironmentSuffix $CondaEnvironmentSuffix
     Invoke-Tests -PackageName $PackageName -EnvName $EnvName
```
anpaz, line 26: This will immediately stop if the tests of a package fail, without testing the rest of the packages.
It's typically better to actually run the tests of all packages so we can catch all failures in the same run.
Also, I'm not certain about the Powershell's semantics and if `$LastExitCode` will report the result of `Invoke-Tests` or the result of the last process (in this case `pytest`). My recommendation would be to change `Invoke-Test` to return $True on success and use that to check for package tests results.

guenp:
It looks like `azure-core` 1.20.0 was just released and is breaking the unit tests. I have pinned to 1.19.1 for now and suggest to fix this issue in a follow-up PR.

build/test.ps1
guenp: Thanks, that sounds good.
One issue I ran into is that if I have `Invoke-Tests` return a boolean, the output of `pytest` no longer shows up in the stdout, and if I pipe it to `Write-Host` instead I can no longer catch the `$LastErrorCode`. Do you have any recommendations on how to fix this issue?

guenp:
@anpaz Thanks for your review! I believe I addressed your comment. The pipeline now fails properly if the tests fail: https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=35163&view=logs&j=2fd4c8cd-b652-5e1a-fd76-c1a09afb8d27&t=ea59e077-c2d1-5af3-f3d7-5c64fa304bd6&l=93

anpaz (approved):

PR #182, opened 2021-11-17
Add conent type to target

sanjgupt:


guenp (approved):
Thanks for catching this!

PR #183, opened 2021-11-18
Add protobuf dependency in requirements

sanjgupt:


azure-quantum/requirements.txt
```diff
@@ -7,3 +7,4 @@ numpy>=1.21.0
 deprecated>=1.2.12
 aiohttp>=3.7.0
 aiofile>=3.7.2
+protobuf==3.14.0
```
guenp, line 10:
```suggestion
protobuf>=3.14.0
```

guenp (changes requested):
Can we use this version as a lower limit?

guenp (approved):

anpaz (approved):

guenp:
Looks like we've got conflicting version numbers:
> The conflict is caused by:
    azure-quantum[cirq,qiskit] 0.21.2111.35696a1 depends on protobuf>=3.14.0
    cirq-google 0.12.0 depends on protobuf~=3.13.0

guenp:
@sanjgupt I've pushed a change to drop `cirq-google` from the requirements (since we really only need `cirq-core`)

PR #187, opened 2021-11-19
chore: update environment.yml

ltalirz:
* Updating environment.yml to python 3.9
 * removing rdkit channel (rdkit is on conda-forge)
 * move further packages from pip to conda
 * give preference to conda-forge channel over quantum-engineering
   (only take packages from quantum-engineering that are missing on
conda-forge)


ltalirz:
It is possible that the last item should be reversed (let me know). My thinking here was that conda-forge is likely to have more up-to-date versions of the various dependencies, so it's better to look there first.

I've tested the `environment.yml` file under ubuntu 20.04, conda 4.10.3, WSL2

qdk/environment.yml
```diff
@@ -1,19 +1,18 @@
 name: qdk
 channels:
-  - rdkit
-  - quantum-engineering
   - conda-forge
+  - quantum-engineering
```
cgranade, line 4: `quantum-engineering` only hosts the `qsharp` and `iqsharp` packages, but has since been deprecated in favor of the `microsoft` channel.
```suggestion
  - microsoft
```

guenp (approved):

PR #194, opened 2021-12-04
Add Target.estimate_price method to calculate job costs for Honeywell and IonQ

### LINKED ISSUE #193: Not clear how to estimate job costs
The costs for submitting jobs to a QPU are currently not transparent. It would be great if there were a way to easily estimate the costs for a circuit.

For Honeywell QPU's, the costs are recorded in "HQCs", Honeywell Quantum Credits, as follows:

`HQC = 5 + C(N_1q + 10 N_2q + 5 N_m)/5000`

where N_1q is the number of 1-qubit gates, N_2q is the number of 2-qubit gates, N_m is the number of state prep and measurement operations, and C is the shot count.

For IonQ QPUs, costs are calculated per gate-shot: the number of gates in the circuit, multiplied by the number of shots.

`cost = ($0.00003 * N_1q + $0.0003 * (N_2q + 6 * (N_multi - 2)) * num_shots`

with a minimum cost of $1
where N_1q is the number of one-qubit gates, N_2q is the number of two-qubit gates, N_multi is the number of qubits involved in a multi-controlled gate, num_shots is the number of shots

guenp:
fixes #193


azure-quantum/azure/quantum/target/target.py
```diff
@@ -158,4 +158,11 @@ def supports_protobuf(self):
...
-        return False
\ No newline at end of file
+        return False
+    
+    def calculate_cost(
+        self,
+        input_data: Any,
+        input_params: Dict[str, Any] = None
+    ):
+        return NotImplementedError("Cost calculation is not implemented for this target")
```
anpaz, line 168:
```suggestion
        return NotImplementedError("Cost calculation is not implemented yet for this target.")
```

azure-quantum/azure/quantum/target/honeywell.py
```diff
@@ -18,6 +18,35 @@ class Honeywell(Target):
         "honeywell.hqs-lt-s1-sim",
     )
 
+    # Get all one-qubit, two-qubit gates and measurement operations
+    GATES_1Q = [
```
anpaz, line 22: what about other 1 qubit gates not listed here?
it might be better to calculate if a gate is 1 or more qubits based on the number of targets.

azure-quantum/azure/quantum/target/ionq.py
```diff
@@ -82,3 +112,68 @@ def submit(
...
+            for one shot, defaults to 0.0003
+        :type cost_2q: float, optional
+        :param min_cost: The minimum cost for running a job, defaults to 1.0
+        :type min_cost: float, optional
+        """
+        if self.name == "ionq.simulator": 
+            return 0.0
+
+        def is_1q_gate(gate: Dict[str, Any]):
+            return gate.get("gate") in self.GATES_1Q and (
```
anpaz, line 157: can we check instead how many target qubits are there in the gate?

guenp: Currently, we're checking both. This list is exhaustive and taken from the IonQ docs: https://docs.ionq.com/#section/Sample-JSON-Circuits

azure-quantum/azure/quantum/target/honeywell.py
guenp: Sure, I'll have to check if the QASM ast contains info about the targets...

azure-quantum/azure/quantum/target/honeywell.py
```diff
@@ -79,3 +108,64 @@ def submit(
...
+    
+    def calculate_cost(
+        self,
+        circuit: str = None,
+        num_shots: int = None,
+        N_1q: int = None,
+        N_2q: int = None,
+        N_m: int  = None
+    ):
+        """Calculate the cost in HQN for a given circuit.
```
Mobius5150, line 120:
```suggestion
        """Calculate the cost in HQC for a given circuit.
```

Mobius5150 (dismissed):
Hey thanks for adding this! I have a few general comments:

1. I'm concerned with calling this "calculate_cost" over "estimate_cost". Did you consider this verbiage and decide against it? "Calculate" sounds like it will definitely cause the job to cost this much, when it can change in the future.

2. Is there an async pattern that could be adapted here? In the future I suspect we'll expose a very similar API that will do estimation on the service. If there's a straightforward way of adapting a forward-compatible interface that could be valuable.

3. More question - does this approach work for cirq + qiskit?

guenp:
@Mobius5150, 1. we went with calculate_cost because it most accurately describes what it does: it calculates the expected cost based on the specified circuit and other input parameters. The docstring clearly explains it's an estimation and the actual billed amount could be different. We will connect this with the pre-submission job cost reporting provided on the service side once it becomes available.
2. This interface is indeed designed to be forward compatible. With async, do you mean async i/o or something else?
3. I discussed with @anpaz and we will add methods to the Qiskit and Cirq abstractions for this feature as well.

guenp:
I had a chat with the Cirq team and they suggested `estimate_price` to be more accurate. `cost` can also be associated with time or other types of resources.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -72,12 +85,12 @@ def run(self, circuit, **kwargs):
...
                 input_params[opt] = kwargs.pop(opt)
 
-        logger.info(f"Submitting new job for backend {self.name()}")
+        logger.info(f"Submitting new job for basckend {self.name()}")
         job = AzureQuantumJob(
             backend=self,
             name=circuit.name,
             target=self.name(),
-            input_data=input_data,
+            input_data=json.dumps(input_data),
```
anpaz, line 93: no, this is wrong. the input_data is an object, not a string.

guenp: Yes you're right, it's a `bytes` object. I've corrected this and am using `IonQ._encode_input_data`.

azure-quantum/azure/quantum/target/honeywell.py
guenp: I looked into this, it looks like [DAGCircuit](https://qiskit.org/documentation/stubs/qiskit.dagcircuit.DAGCircuit.html) has the method `DAGCircuit.op_nodes` that returns all the DAG nodes that are operations (`DAGOpNode`) which has a property `op` that includes the instruction name, number of qubits and number of classical bits. E.g.:
```python
print([n.op for n in dag.op_nodes()])
[Instruction(name='h', num_qubits=1, num_clbits=0, params=[]), Instruction(name='cx', num_qubits=2, num_clbits=0, params=[]), Instruction(name='cx', num_qubits=2, num_clbits=0, params=[]), Instruction(name='measure', num_qubits=1, num_clbits=1, params=[]), Instruction(name='measure', num_qubits=1, num_clbits=1, params=[]), Instruction(name='measure', num_qubits=1, num_clbits=1, params=[])]
```
I could find the number of gates of each type by filtering `dag.op_nodes()` by `num_qubits`, and pick out the `measure` and `reset` operations by name. Would something like that work?

anpaz (approved):

anpaz (approved):
Please make sure we update the code as soon as we can obtain this from the service.

PR #195, opened 2021-12-11
Update 1Qbit solver parameter

KatieO4t1qbit:
Added seed param to the 1qbit solvers and removed perform_icm from pticm


guenp:
Thanks @KatieO4t1qbit! Since this is a breaking change, could you share a bit more about whether these changes are backward compatible? (i.e. what happens if `perform_icm` is specified, does the solver throw an error or is it simply ignored)? We'll then make sure to include this in the release notes.

KatieO4t1qbit:
Hi @guenp 
An error will be thrown if the `perform_icm` field is specified. I'm not sure where to update the release notes, can you please point me to them?
I'll take a look at the unit tests, but at a glance they don't seem to be specifying `perform_icm` anywhere.

KatieO4t1qbit:
The tests are passing.

guenp:
@KatieO4t1qbit the release notes are here: https://github.com/MicrosoftDocs/quantum-docs/blob/c57a1cdc4d553163f8469a80b1d30bc67ca3143c/articles/release-notes.md

Your change will be part of the January release. We typically open a PR to that repo when we start the monthly release process. If you'd like to draft a note on the change on this PR then I will make sure to add it there next month.

guenp:
FYI @KatieO4t1qbit I just tested locally and the unit tests are not passing for me. Could you check if you set the `AZURE_QUANTUM_1QBIT` env variable?

anpaz (approved):

PR #199, opened 2021-12-15
Fix bug in parsing IonQ histogram results into a qiskit.result.Result object

### LINKED ISSUE #198: Bug: Qiskit results are overwritten when using helper qubits
The current implementation of `azure.quantum.qiskit` contains a bug: if not all qubits are measured, the resulting histogram counts are incorrect.

The bug is in [this](https://github.com/microsoft/qdk-python/blob/31664c9410b4d7b65e576582fef421ca53651ed7/azure-quantum/azure/quantum/qiskit/job.py#L158) line.

This causes for the results returned for different bitstrings that have the same readout in common for the measured qubits to get overwritten, which means that the total number of counts is incorrect.

### Steps to reproduce

```python
circuit = QuantumCircuit(4, 3)
circuit.name = "Qiskit Sample - 3-qubit GHZ circuit"
circuit.h(0)
circuit.cx(0, 1)
circuit.cx(1, 2)
circuit.h(3) # Dummy helper qubit that is not measured
circuit.measure([0,1,2], [0, 1, 2])

provider = AzureQuantumProvider(...)
job = backend.run(circuit=circuit, shots=1e3)
result = job.result()

assert assert result.data()["counts"] == {
    '000': 500, '111': 500
}
```

This raises:
```stdout
E               AssertionError: assert {'000': 125, '111': 125} == {'000': 250, '111': 250}
E                 Differing items:
E                 {'000': 125} != {'000': 250}
E                 {'111': 125} != {'111': 250}
```

guenp:
Fixes #198

### DIFF
--- a/azure-quantum/tests/unit/test_plugins.py
+++ b/azure-quantum/tests/unit/test_plugins.py
@@ -29,11 +29,12 @@ class TestQiskit(QuantumTestBase):
                else Job.create_job_id()
     
     def _3_qubit_ghz(self):
-        circuit = QuantumCircuit(3, 3)
+        circuit = QuantumCircuit(4, 3)
         circuit.name = "Qiskit Sample - 3-qubit GHZ circuit"
         circuit.h(0)
         circuit.cx(0, 1)
         circuit.cx(1, 2)
+        circuit.h(3) # Dummy helper qubit that is not measured
         circuit.measure([0,1,2], [0, 1, 2])
 
         return circuit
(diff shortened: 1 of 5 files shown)


azure-quantum/tests/unit/test_plugins.py
```diff
@@ -29,11 +29,12 @@ def get_test_job_id(self):
                else Job.create_job_id()
     
     def _3_qubit_ghz(self):
```
anpaz, line 31: I think we still need to add a test that confirms that the counts in the `result` object match the expectation, or more broadly that the results returned from the service are correctly parsed.

anpaz (approved):

PR #222, opened 2022-01-25
update recording and test for latest Honeywell API validator changes

guenp:
The API validator now returns a single digit instead of a bitstring. This PR updates the unit test and recording.


anpaz (approved):

guenp:
Merging blocked by Katas e2e build bug, fixed here: https://github.com/microsoft/QuantumKatas/pull/737

PR #232, opened 2022-02-05
Test pipeline improvements

guenp:
Some improvements to our tests and CI/CD:

- Added a more robust workaround option for https://github.com/microsoft/qdk-python/issues/118 such that we don't have to manually edit test recordings.
- Disable qdk-python.e2e pipeline on PRs to reduce churn when unrelated packages in the QDK are failing.
- Added note to CONTRIBUTING.md to remind folks to update any relevant test recordings before merging to `main`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,6 +53,12 @@ To run the a specific unit test case, run:
 pytest -k test_job_refresh
 ```
 
+#### Recordings ####
+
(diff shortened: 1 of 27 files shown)


guenp:
Pushed a fix for another issue that incidentally showed up - the latest `boost-cpp` build seems to be incompatible with the `rdkit` build on conda-forge. Fixing to the previous build of `boost-cpp=1.74.0` seems to have fixed the issue.
I posted more details on the issue here: https://github.com/rdkit/rdkit/issues/4987

vxfield (approved):
Thanks for the additions.
Let's see if in the future we can move the mock of wait_complete to the test base class.

PR #240, opened 2022-02-08
Fix issue with use_hardware flag

guenp:


azure-quantum/azure/quantum/projectq/backends/honeywell.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         workspace.append_user_agent(PROJECTQ_USER_AGENT)
         self._workspace = workspace
 
-        if not use_hardware:
+        if not use_hardware and "apival" not in device:
```
anpaz, line 63: I think a better logic here is:
1. let `device` be None by default.
2. check if the user has explicitly set a device simply use that (shouldn't this be called `target_id` to be consistent with the rest of the API, or is this a well known parameter in projectq?)
3. otherwise, select a default device based on the `use_hardware` parameter.

We should add a unittest that verifies this logic is implemented correctly.

anpaz (approved):

azure-quantum/azure/quantum/projectq/backends/honeywell.py
guenp: Done

PR #241, opened 2022-02-09
remove compress flag from solvers.py as it is no longer required for …

sanjgupt:
This PR removes the compress flag from the solvers submit method as this is no longer required to support toshiba. 
Please see this request: https://github.com/microsoft/qdk-python/issues/235


guenp:
Thanks @sanjgupt! Since the behavior of the target changed, I think we should re-record the Toshiba tests.

guenp (approved):
LGTM, thanks @sanjgupt!

PR #244, opened 2022-02-10
Roll back ProjectQ support + deprecate azure.quantum.plugins

guenp:
This feature needs a bit more work to make it stable. I migrated the code back to a feature branch here: https://github.com/microsoft/qdk-python/tree/feature/projectq


anpaz (approved):
The only comment is that we should consider breaking plugin tests to different files, such that adding/removing frameworks and/or partners is self-contained.

PR #245, opened 2022-02-11
Improve error msg for missing SKU

guenp:


guenp:
This PR depends on a docs update.

azure-quantum/azure/quantum/cirq/service.py
```diff
@@ -137,11 +137,16 @@ def create_job(
         :rtype: azure.quantum.cirq.Job
         """
         # Get target
-        target = self.get_target(name=target)
+        _target = self.get_target(name=target)
+        if not _target:
+            target_name = target or self._default_target
+            raise RuntimeError(f"Could not find target '{target_name}'. \
```
xinyi-joffre, line 143: Is it possible to print the provider name in the error message as well? Users may not know which provider to add based on the missing target, although I guess if they looked closely, all targets have the provider the dot?

azure-quantum/azure/quantum/qiskit/provider.py
```diff
@@ -32,6 +33,19 @@ def __init__(self, workspace=None, **kwargs):
     def get_workspace(self) -> Workspace:
         return self._workspace
 
+    def get_backend(self, name=None, **kwargs):
+        """
+        Return a single backend matching the specified filtering.
+        """
+        backends = self.backends(name, **kwargs)
+        if len(backends) > 1:
+            raise QiskitBackendNotFoundError("More than one backend matches the criteria")
```
xinyi-joffre, line 42: What does it mean when more than one backend matches the criteria? Are there next steps we can suggest to the user?

azure-quantum/azure/quantum/cirq/service.py
xinyi-joffre: Maybe we can say "Could not find target '{target_name}' for provider '{parsed_provider_id}."

Is there also a possibility that the provider is provisioned, but the target is not supported in the region? This sometimes happens with QIO, where FPGA is not supported in all regions. In those cases, will the target also be missing?

xinyi-joffre (approved):

azure-quantum/azure/quantum/qiskit/provider.py
guenp: This is unlikely to happen, I included this code here because it mimics the behavior of the parent class (Qiskit backend).

azure-quantum/azure/quantum/target/honeywell.py
```diff
@@ -95,7 +95,7 @@ def estimate_cost(
         The actual price charged by the provider may differ from this estimation.
 
         For the most current pricing details, see
-        https://docs.microsoft.com/en-us/azure/quantum/provider-honeywell#honeywell-system-model-h1
+        https://docs.microsoft.com/azure/quantum/provider-honeywell#honeywell-system-model-h1
```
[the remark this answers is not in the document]
guenp: Ah you're right, it should be `https://docs.microsoft.com/azure/quantum/provider-honeywell#system-model-h1-powered-by-honeywell`. Thanks for pointing that out.

azure-quantum/azure/quantum/cirq/service.py
guenp: Currently, it's possible to get target only from the name, so we don't require a provider ID here. So yes, in that case the target will also be missing.

guenp:
Thanks @xinyi-joffre for reviewing! I've opened a new PR to address your comments (this one was already merged before I could address them, because I had turned on automerge).

PR #246, opened 2022-02-11
Make Honeywell/Quantinuum backend consistent with shots input argument

guenp:
`backend.run` and `backend.estimate_cost` should take `shots` as input argument; for the Honeywell backend, this is currently `count`.
This PR changes the input argument to `shots` and adds a deprecation warning for `count`.

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/backends/honeywell.py
+++ b/azure-quantum/azure/quantum/qiskit/backends/honeywell.py
@@ -2,6 +2,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 ##
+import warnings
+
 from typing import TYPE_CHECKING, Union, List
 from azure.quantum.version import __version__
 from azure.quantum.qiskit.job import AzureQuantumJob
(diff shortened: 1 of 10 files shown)


azure-quantum/azure/quantum/qiskit/backends/honeywell.py
```diff
@@ -60,23 +62,36 @@ class HoneywellBackend(Backend):
...
 
         :param circuit: Qiskit quantum circuit
         :type circuit: QuantumCircuit
-        :param count: Shot count
+        :param shots: Shot count
+        :type shots: int
+        :param count: Shot count (deprecated)
         :type count: int
         """
+        if shots is None or count is not None:
```
vxfield, line 77: I recommend minimizing the use of "count", and just handling it once at the beginning for backward compatibility.
After this, only use "shots"
```suggestion
        if count is not None:
            shots = count
            warnings.warn(
                "Input parameter 'count' has been deprecated. Please use 'shots' instead.")
```

azure-quantum/azure/quantum/qiskit/backends/honeywell.py
```diff
@@ -60,23 +62,36 @@ class HoneywellBackend(Backend):
...
+            shots = shots or count
+            warnings.warn(
+                "Input parameter 'count' has been deprecated. Please use 'shots' instead.")
+        if shots is None and count is None:
```
vxfield, line 81:
```suggestion
        if shots is None:
```

azure-quantum/azure/quantum/qiskit/backends/honeywell.py
```diff
@@ -60,23 +62,36 @@ class HoneywellBackend(Backend):
...
+
         input_data = circuit.qasm()
         workspace = self.provider().get_workspace()
         target = workspace.get_targets(self.name())
-        return target.estimate_cost(input_data, num_shots=count)
+        return target.estimate_cost(input_data, num_shots=shots)
 
     def run(self, circuit: Union[QuantumCircuit, List[QuantumCircuit]], **kwargs):
         """Submits the given circuit for execution on a Honeywell target."""
+        if "count" in kwargs and "shots" not in kwargs:
```
vxfield, line 91:
```suggestion
        if "count" in kwargs:
```

vxfield (changes requested):
It would be good to add unit test for the change, covering the new and old parameter name.

vxfield (approved):

PR #249, opened 2022-02-14
Cost estimate for Honeywell simulator should be zero

guenp:


guenp:
Closing this since the simulator does cost credits (EHQC). We will pick this up in the documentation.

PR #250, opened 2022-02-14
Fix issue with simulator counts

### LINKED ISSUE #248: Number of total shots is wrong after processing results in Qiskit plugin
When submitting a Qiskit program, the total number of shots does not match the shots requested.

```python
provider = AzureQuantumProvider(...)
num_shots = 1000

job = backend.run(
    circuit=self._5_qubit_superposition(),
    shots=num_shots
)

result = job.result()
assert result.data()["counts"]
```

```bash
{'0': 496, '1': 496}
```

However, it looks like the actual histogram returned by the provider does add up to 1:

```python
result = job._azure_job.get_results()
sum(result["histogram"].values())
```

```bash
1.0
```

It looks like this is potentially caused by an integer rounding error in `azure.quantum.qiskit.AzureQuantumJob._format_ionq_results`.

guenp:
Fixes #248 

Some open questions:
- There's currently not a way to check if a target is a simulator other than the name; should there be a better way?
- This is a similar implementation to how `qiskit_ionq` is implemented, would it be worth refactoring to use it as a dependency?
- Should `job.result()` take a `sampler_seed` or should it be fetched from the metadata (similar to `qiskit_ionq`)?

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/job.py
+++ b/azure-quantum/azure/quantum/qiskit/job.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 ##
 from collections import defaultdict
+import numpy as np
 
 try:
     from qiskit.providers import JobV1, JobStatus
@@ -71,12 +72,12 @@ class AzureQuantumJob(JobV1):
         self._azure_job.submit()
         return
 
-    def result(self, timeout=None):
+    def result(self, timeout=None, sampler_seed=None):
         """Return the results of the job."""
         self._azure_job.wait_until_completed(timeout_secs=timeout)
 
         success = self._azure_job.details.status == "Succeeded"
-        results = self._format_results()
+        results = self._format_results(sampler_seed=sampler_seed)
 
         return Result.from_dict(
             {
@@ -106,7 +107,7 @@ class AzureQuantumJob(JobV1):
         """Return the position of the job in the queue. Currently not supported."""
         return None
 
-    def _format_results(self):
+    def _format_results(self, sampler_seed=None):
         """ Populates the results datastructures in a format that is compatible with qiskit libraries. """
         success = self._azure_job.details.status == "Succeeded"
         shots_key = "shots"
@@ -118,12 +119,13 @@ class AzureQuantumJob(JobV1):
         }
 
         if success:
+            is_simulator = "sim" in self._azure_job.details.target
             if (self._azure_job.details.output_data_format == MICROSOFT_OUTPUT_DATA_FORMAT):
-                job_result["data"] = self._format_microsoft_results()
+                job_result["data"] = self._format_microsoft_results(sampler_seed=sampler_seed)
                 job_result["header"] = { "name": self._azure_job.details.name }
                 
             elif (self._azure_job.details.output_data_format == IONQ_OUTPUT_DATA_FORMAT):
-                job_result["data"] = self._format_ionq_results()
+                job_result["data"] = self._format_ionq_results(sampler_seed=sampler_seed, is_simulator=is_simulator)
                 job_result["header"] = self._azure_job.details.metadata
 
             elif (self._azure_job.details.output_data_format == HONEYWELL_OUTPUT_DATA_FORMAT):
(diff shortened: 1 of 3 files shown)


azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -135,41 +137,61 @@ def _format_results(self):
...
-        shots = self._azure_job.details.input_params[shots_key] if shots_key in self._azure_job.details.input_params else self._backend.options.get(shots_key)
+        shots = self._azure_job.details.input_params[shots_key] \
+            if shots_key in self._azure_job.details.input_params \
+            else self._backend.options.get(shots_key)
         job_result["shots"] = shots
         return job_result
 
-    def _format_ionq_results(self):
+    @staticmethod
+    def _draw_random_sample(sampler_seed, probabilities, shots):
```
anpaz, line 147: is this drawing a random sample? looks more like it's calculating a bunch of values instead of just drawing one?

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -135,41 +137,61 @@ def _format_results(self):
...
+        rand = np.random.RandomState(sampler_seed)
```
anpaz, line 148: When you do this, normally this means that all the random numbers will return the same values, so calling the same method multiple times will return the same value (as such, not really drawing a random sample). Is this really the purpose?

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 ##
 from collections import defaultdict
+import numpy as np
```
anpaz, line 6: Are we adding `numpy` as a dependency just for this? Since a bit heavy just to generate a random number.

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -71,12 +72,12 @@ def submit(self):
...
-    def result(self, timeout=None):
+    def result(self, timeout=None, sampler_seed=None):
         """Return the results of the job."""
         self._azure_job.wait_until_completed(timeout_secs=timeout)
 
         success = self._azure_job.details.status == "Succeeded"
-        results = self._format_results()
+        results = self._format_results(sampler_seed=sampler_seed)
 
         return Result.from_dict(
```
anpaz, line 82: Can we cache the results of this call as a field of this object, so we don't have to download and parse the results everytime the user call `result()`?

anpaz (commented):
I don't know why we would need to generate different counts for the simulator vs the qpu, as such I don't think we need to differentiate them. If counts are missing from both, why do we need different ways to calculate counts for each?

If the dependency on the ionq package is already there, then I'd be fine to just reuse their code, otherwise I'd rather not introduce that dependency.

If we want the counts to be consistent then, yes, the seed should go in the metadata so we use the same on every output.

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -182,13 +204,17 @@ def _format_microsoft_results(self):
...
-                key = histogram[i]
+                bitstring = histogram[i]
                 value = histogram[i + 1]
-                probabilities[key] = value
-                counts[key] = int(shots * value)
+                probabilities[bitstring] = value
         else:
             raise "Invalid number of items in Job results' histogram."
 
+        if is_simulator:
```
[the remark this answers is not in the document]
guenp: The QPUs actually generate shot counts while the simulator emulates them. Shouldn't we return the actual shots produced by the QPU?

azure-quantum/azure/quantum/qiskit/job.py
guenp: In statistics, a "sample" typically refers to a set of data points, see: https://en.wikipedia.org/wiki/Sample_(statistics)

azure-quantum/azure/quantum/qiskit/job.py
guenp: Yes, that is the purpose. If I specify a finite `sampler_seed` I want this function to always return the same values. This can be helpful for testing purposes.

azure-quantum/azure/quantum/qiskit/job.py
guenp: This is actually already the behavior of `Job.get_results` which is called under the hood by `_format_results`.

vxfield (approved):

PR #261, opened 2022-02-25
Override SimulatedAnnealing factory method to set platform

### LINKED ISSUE #260: Simulated annealing parameter free target appears twice when querying workspace targets
image

As you can see in the image, "microsoft.simulatedannealing-parameterfree.cpu" target appears twice.

Debugging the code; the service is returning 2 distinct targets:
"microsoft.simulatedannealing-parameterfree.cpu"
"microsoft.simulatedannealing-parameterfree.fpga"

But due to [this code](https://github.com/microsoft/qdk-python/blob/81b42e50cffeb1e438242e790d685b40b75e8fca/azure-quantum/azure/quantum/target/microsoft/qio/simulated_annealing.py#L67), the fpga target gets converted to cpu target.

guenp:
Fixes #260

### DIFF
--- a/azure-quantum/azure/quantum/target/microsoft/qio/simulated_annealing.py
+++ b/azure-quantum/azure/quantum/target/microsoft/qio/simulated_annealing.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 ##
 import logging
+from multiprocessing.sharedctypes import Value
 
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


anpaz (approved):
Should we add some unittest for this?

george-moussa (approved):

guenp:
@anpaz, we should but we don't have an FPGA solver enabled in our test workspace at the moment. We could either add it or create a non-live recorded test against a workspace that already does.

PR #268, opened 2022-03-09
Add Rigetti Target

dbanty:
Adds the `Rigetti` target along with some helper classes for processing results. Here's a little script of how I expect folks will use these:

```python
from azure.quantum.target.rigetti import Result, Rigetti, RigettiTarget

workspace = Workspace(
    # TODO: Fill in your details here
    subscription_id="",
    resource_group="",
    name="",
    location=""
)

target = Rigetti(
    workspace=workspace,
    name=RigettiTarget.ASPEN_11,  # Defaults to RigettiTarget.QVM for simulation
)

# Any valid Quil program is accepted, but the readout must be named `ro`
readout = "ro"
bell_state_quil = f"""
DECLARE {readout} BIT[2]

H 0
CNOT 0 1

MEASURE 0 {readout}[0]
MEASURE 1 {readout}[1]
"""

num_shots = 5
job = target.submit(
    input_data=bell_state_quil, 
    name="bell state", 
    input_params={"count": num_shots},
)

print(f"Queued job: {job.id}")
job.wait_until_completed()

print(f"Job completed with state: {job.details.status}")
result = Result(job)  # This throws an exception if the job failed
# You can index a Result with the name of the readout. In this case, `ro`
data_per_shot = result[readout]
# Here, data_per_shot is a list of length num_shots, each entry is a list containing the data for the register for that shot
ro_data_first_shot = data_per_shot[0]
# In this case, because the type of the register is BIT, the type will be integer and the value either 0 or 1
assert isinstance(ro_data_first_shot[0], int)
assert ro_data_first_shot[0] == 1 or ro_data_first_shot[0] == 0

# Let's print out all the data
print(f"Data from '{readout}' register:")
for i, shot in enumerate(data_per_shot):
    print(f"Shot {i}: {shot}")
```


guenp (changes requested):
Thanks @dbanty for implementing this new target!
Before merging to `main` we need to add unit tests. You can find instructions on how to do that here: https://github.com/microsoft/qdk-python/tree/main/azure-quantum/tests. I suggest to add a new test file `azure-quantum/tests/unit/test_rigetti.py`. Let me know if I can help with this step or if you have any questions.

dbanty:
Thanks for the feedback @guenp! I made the requested changes and the tests pass for me locally. I assume they're failing in CI because the runner doesn't have access to the target—should I be committing the recording as well or do we want to wait until we can get CI hooked up to Rigetti targets?

dbanty:
With the recording checked in it looks like CI is passing 🥳. I also got approval for the CLA, so that's signed. Please let me know if there's anything else needed to move forward with this. Thanks again for the feedback!

guenp (dismissed):
Thank you @dbanty, this looks great. @ricardo-espinoza, could you please take another look at the PR, make sure that our internal e2e tests are succeeding and let Dylan know when it's ready to merge?

dbanty:
Hey, just a heads up that we're going to be renaming the targets per today's sync meeting, as well as adding some types for expected inputParams. I'll let you know once we're ready for re-review.

george-moussa:
@dbanty backed changes are deployed, let us know when this is ready for merge.

dbanty:
@george-moussa and @guenp the targets have all been renamed and are confirmed working (with new recorded tests). I'll add the typed input parameters in a follow-up PR instead of this one for expediency as we still have more validation to do for those.

ricardo-espinoza (approved):

george-moussa (approved):

PR #273, opened 2022-03-21
Anatoliy/switch tabu to qiote

Anatoliy-Litvinenko:
Shipment of new version of Tabu solver for preview, preserving old Tabu solver.


sanjgupt:
Hi please add a unit test to verify that this switch works correctly in different scenarios, i.e. when 'experimental' is not in name vs it is, the correct code path is selected.

PR #277, opened 2022-03-25
Propagate circuit metadata to Azure Quantum job for IonQ backend

ricardo-espinoza:
Running StateTomography from qiskit_experiments requires information from the circuit metadata that should be present in the job output. If those properties are not present, analysis of results will fail.

This change adds the corresponding circuit metadata.


ricardo-espinoza:
This pull request has been replaced by #278

Issue #263, opened 2022-02-28
Missing method retrieve_job on Honeywell/Quantinuum backend


guenp:
For ionq qpu, I used the following code block 

 
```python
simulator = myprovider.get_backend('ionq.qpu')

retrieved =simulator.retrieve_job("myid")
```
 

But similar code block does not work with Honeywell 

```python
simulator = provider.get_backend("honeywell.hqs-lt-s2")

retrieved =simulator.retrieve_job("myid")
```
 

It results in the following error message

```python
AttributeError: 'HoneywellQPUBackend' object has no attribute 'retrieve_job'
```

ricardo-espinoza:
With the changes in #278 , the method `retrieve_job()` is now part of the base implementation and is available to all providers.

Issue #259, opened 2022-02-08
Qiskit Job Id Retrival Error for qiskit-QAOA with Ionq Azure Quantum Provider


leonox:
**Describe the bug**
Dear Azure Quantum Devs&Users, I am currently using qiskit on ionq hardware via azure quantum providers (as part of a quantum inspired application). Using the std example I can directly run circuits but not a qaoa optimization (Error:"FAILURE: Can not get job id, Resubmit the qobj to get job id. Error: 'list' object has no attribute 'clbits' ") - could this be because Quantum Instance does not utilizie ionq_provider/backend yet or the azure provider is missing arguments and what could help to identify/resolve the issue?

**To Reproduce**
Code excerpts:
`azure_provider = AzureQuantumProvider(subscription_id="subid",
                              resource_group="rog",
                              name="qc",
                              location="West Europe",
                              credential=credential)
simulator_backend = azure_provider.get_backend('ionq.simulator')
seed = 42
cobyla = COBYLA()
cobyla.set_options(maxiter=250)
quantum_instance = QuantumInstance(backend=simulator_backend, seed_simulator=seed, seed_transpiler=seed)
qaoa_mes = QAOA(optimizer=cobyla, reps=3, quantum_instance=quantum_instance)
qaoa = MinimumEigenOptimizer(qaoa_mes)
result = qaoa.solve(qp)` throws the error
with `simulator_backend = Aer.get_backend('aer_simulator')` it runs like a charm

`azure_provider = AzureQuantumProvider(subscription_id="subid",
                              resource_group="rog",
                              name="qc",
                              location="West Europe",
                              credential=credential) 
simulator_backend = azure_provider.get_backend('ionq.simulator');
quantum_instance = QuantumInstance(backend=simulator_backend, seed_simulator=seed, seed_transpiler=seed)
circuit = QuantumCircuit(3, 3)
circuit.name = "Qiskit Sample - 3-qubit GHZ circuit"
circuit.h(0)
circuit.measure([0, 1, 2], [0, 1, 2])
job = simulator_backend.run(circuit, shots=100)
result = job.result()` works

**Expected behavior**
return result without error

**System information**
 - OS: Windows
- packages:
 azure-quantum             0.21.2112.180703          pypi_0    pypi
 qiskit                    0.34.1                   pypi_0    pypi
qiskit-aer                0.10.2                   pypi_0    pypi
qiskit-ibmq-provider      0.18.3                   pypi_0    pypi
[...]

Issue #258, opened 2022-02-24
Test function needs to be async


IsraelMiles:
A Python test is marked as `@pytest.mark.asyncio` but is not async. See here:
[qdk-python/test_aio_solvers.py at 07c3ea31d9e53aac154589f54a9ada6482bd3242 · microsoft/qdk-python (github.com)](https://github.com/microsoft/qdk-python/blob/07c3ea31d9e53aac154589f54a9ada6482bd3242/azure-quantum/tests/unit/aio/test_aio_solvers.py)

@guenp

anpaz:
Done.

Issue #224, opened 2022-01-27
Cannot submit list of circuits via AzureQuantumBackend.run


guenp:
Currently, to submit multiple circuits, users have to submit each circuit individually, e.g.
```python
jobs = []
for circuit in circuits:
    jobs.append(backend.run(circuit, shots=N))

results = []
for job in jobs:
    results.append(job.result())
```

Iterative algorithms require being able to submit a batch of circuits at the same time. There are several Qiskit libraries that depend on support for `backend.run(circuits: List[QuantumCircuit])`, such as qiskit-experiments and qiskit-aqua. See also: https://qiskit.org/documentation/stubs/qiskit.providers.ibmq.managed.IBMQJobManager.html#qiskit.providers.ibmq.managed.IBMQJobManager

This would require being able to submit multiple circuits and get a single job ID back to fetch the results, similar to the IBMQ experience. However, this is currently not something our service supports; Azure Quantum's Backend currently only accepts a single circuit or a list of circuits of length one, see this line: [ionq.py#L70](https://github.com/microsoft/qdk-python/blob/b75676967a952763bd565b84ddc983de8d25809d/azure-quantum/azure/quantum/qiskit/backends/ionq.py#L70).

For instance, [Tomography](https://github.com/Qiskit/qiskit-experiments/blob/8f28301e8ef59e34569c1fe983f90c7221baa5e5/qiskit_experiments/library/tomography/tomography_experiment.py#L28) uses the `backend.run()` method under the hood which returns a single job: [base_experiment.py#L345](https://github.com/Qiskit/qiskit-experiments/blob/8f28301e8ef59e34569c1fe983f90c7221baa5e5/qiskit_experiments/framework/base_experiment.py#L345).

guenp:
It is currently not possible to use e.g. `VQEProgram` because `backend.run(circuit)` currently only accepts a single circuit, not a list of circuits:
![error_screen](https://user-images.githubusercontent.com/4041805/151723170-bb09a5c3-4a0f-4a99-82d4-276baa50c712.png)

anpaz:
@guenp
Adding support for job batching is a great idea, and it's on our road-map, but to be perfectly clear it is **not** a pre-requisite for qiskit-experiments, QAOA or VQE.

[...]

PR #278, opened 2022-03-25
Propagate Qiskit's circuit metadata to jobs for Azure Quantum backends

ricardo-espinoza:
Running StateTomography from qiskit_experiments requires information from the circuit metadata that should be present in the job output. If those properties are not present, analysis of results will fail.

This change adds the corresponding circuit metadata.

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/backends/backend.py
+++ b/azure-quantum/azure/quantum/qiskit/backends/backend.py
@@ -1,0 +1,112 @@
+##
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+##
+import warnings
+import json
+
+import logging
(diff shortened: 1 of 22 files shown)


anpaz (changes requested):
Can we add/modify some test that checks for this?

anpaz:
This PR should address #259

anpaz:
This is also a fix for running VQE, mentioned in #224

anpaz (approved):
The original changes only fixed IonQ backends. My latest push refactors the Qiskit backends to have a common base class that appends the metadata by default. It also removes a bunch of duplicated code.

anpaz:
The new common base Backend class addresses #263

anpaz:
addresses #258

PR #279, opened 2022-03-29
feat: Add typed `InputParams` to document available params for Rigetti targets.

dbanty:
The goal was to explicitly document the availability of the `skipQuilc` parameter in a way that is easily discoverable (e.g., by inspecting the signature of `submit`).

If there's a more recommended pattern than using a class for this, I'm happy to swap things around!


anpaz (approved):
I would suggest adding another submit test that takes no input_params, to make sure the defaults work correctly.

dbanty:
Thanks for the suggestion @anpaz, I went ahead and added some additional test cases for `input_params`.

anpaz (approved):

anpaz:
@dbanty, i've approved and trying to merge but it says some checks are still missing. I believe it's because your branch is a bit behind of `main`. Can you please try updating?

dbanty:
Updated!

PR #295, opened 2022-04-18
Use QIR as the default input-data-format from Qiskit

anpaz:
Adding support to submit circuits using QIR as the datastructure, if the input-data-format is marked as "qir.v1".


ricardo-espinoza (approved):

raghavthind2005 (approved):

azure-quantum/setup.py
```diff
@@ -82,7 +82,8 @@
     extras_require={
         "qiskit": [
             "qiskit-ionq>=0.1.4",
-            "qiskit-terra>=0.19.1"
+            "qiskit-terra>=0.19.1",
+            "qiskit-qir"
```
ricardo-espinoza, line 86: We'll need to specify here the version that includes the recent changes on the qiskit-qir package.

PR #298, opened 2022-04-19
QIR support for Rigetti

anpaz:


azure-quantum/azure/quantum/qiskit/backends/rigetti.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        logger.info("Initializing RigettiSimulatorBackend")
+        configuration: BackendConfiguration = kwargs.pop("configuration", default_config)
+        super().__init__(configuration=configuration, provider=provider, **kwargs)
+
+
+class RigettiQPUBackend(RigettiBackend):
+    backend_names = (
+        RigettiTarget.ASPEN_11.value,
+        RigettiTarget.ASPEN_M_1.value
```
dbanty, line 86: Should we maybe make a helper function on `RigettiTarget` called `qpus()` or similar to make this easier to keep up to date?

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
```diff
@@ -0,0 +1,117 @@
...
+            {
+                "backend_name": name,
+                "backend_version": __version__,
+                "simulator": False,
+                "local": False,
+                "coupling_map": None,
+                "description": "Rigetti QPU on Azure Quantum",
+                "basis_gates": QIR_BASIS_GATES,
+                "memory": False,
+                "n_qubits": 80 if name == RigettiTarget.ASPEN_M_1.value else 38,
```
dbanty, line 106: Perhaps this should also be a function or similar on `RigettiTarget` so it's more clear that it needs to be updated when new QPUs are added.

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
```diff
@@ -0,0 +1,117 @@
...
+            {
+                "backend_name": name,
+                "backend_version": __version__,
+                "simulator": True,
+                "local": False,
+                "coupling_map": None,
+                "description": "Rigetti simulator on Azure Quantum",
+                "basis_gates": QIR_BASIS_GATES,
+                "memory": False,
+                "n_qubits": 38,
```
dbanty, line 69: Is this a limit on the number of qubits that can be simulated? If so, I think we have to lower it based on QVM's usage. I can get you an exact number with a bit of digging but I think it will be closer to 20.

dbanty (commented):
This generally looks good to me, just a couple of questions asked.

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
anpaz: great idea.

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
anpaz: ok. lowered it to 20.

ricardo-espinoza (approved):

azure-quantum/azure/quantum/target/rigetti/target.py
```diff
@@ -32,6 +32,30 @@ class RigettiTarget(str, Enum):
...
+        return [
+            RigettiTarget.QVM.value,
+        ]
+
+    def qpus():
+        """Returns a list of QPU targets"""
+        return [
+            RigettiTarget.ASPEN_11.value,
+            RigettiTarget.ASPEN_M_1.value,
+        ]
```
dbanty, line 46:
```suggestion
    def simulators() -> List[str]:
        """Returns a list of simulator targets"""
        return [
            RigettiTarget.QVM.value,
        ]

    def qpus() -> List[str]:
        """Returns a list of QPU targets"""
        return [
            RigettiTarget.ASPEN_11.value,
            RigettiTarget.ASPEN_M_1.value,
        ]
```

Does it matter that these are now lists when before I believe they were tuples?

PR #301, opened 2022-04-22
[azure-quantum] Normalize histogram probabilities if sum close to 1

### LINKED ISSUE #287: Azure Quantum Qiskit Rand.Choice ValueError: probabilities do not sum to 1
This is a follow-up to https://github.com/microsoft/qdk-python/issues/259 
Backends are responsive but the ionq.simulator returns probabilities not summing to 1 creating a value error with numpy`s rand.choice used in the QAOA/Minimumeigensolver of qiskit.

Call stack:
`  File "qiskit/algorithms/minimum_eigen_solvers/vqe.py", line 624, in energy_evaluation
    sampled_expect_op = self._circuit_sampler.convert(expect_op, params=param_bindings)
  File "qiskit/opflow/converters/circuit_sampler.py", line 218, in convert
    sampled_statefn_dicts = self.sample_circuits(circuit_sfns=circs, param_bindings=p_b)
  File "/qiskit/opflow/converters/circuit_sampler.py", line 337, in sample_circuits
    results = self.quantum_instance.execute(
  File "qiskit/utils/quantum_instance.py", line 790, in execute
    run_circuits(
  File "qiskit/utils/run_circuits.py", line 623, in run_circuits
    results.append(job.result())
  File "azure/quantum/qiskit/job.py", line 80, in result
    results = self._format_results(sampler_seed=sampler_seed)
  File "azure/quantum/qiskit/job.py", line 128, in _format_results
    job_result["data"] = self._format_ionq_results(sampler_seed=sampler_seed, is_simulator=is_simulator)
  File "azure/quantum/qiskit/job.py", line 185, in _format_ionq_results
    counts = self._draw_random_sample(sampler_seed, probabilities, shots)
  File "azure/quantum/qiskit/job.py", line 151, in _draw_random_sample
    rand_values = rand.choice(list(probabilities.keys()), shots, p=list(probabilities.values()))
  File "mtrand.pyx", line 939, in numpy.random.mtrand.RandomState.choice
ValueError: probabilities do not sum to 1`

MacOS
python ==3.9
azure-quantum[qiskit]==0.23.201228b1

guenp:
Fixes #287 and https://github.com/microsoft/qdk-python/issues/302
See https://github.com/Qiskit-Partners/qiskit-ionq/blob/306f5c08aab74dbc3b7755444b0f108cb068aebc/qiskit_ionq/ionq_job.py#L98 for similar `qiskit-ionq` implementation

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/job.py
+++ b/azure-quantum/azure/quantum/qiskit/job.py
@@ -147,7 +147,7 @@ class AzureQuantumJob(JobV1):
     def _draw_random_sample(sampler_seed, probabilities, shots):
         _norm = sum(probabilities.values())
         if _norm != 1:
-            if np.isclose(_norm, 1.0):
+            if np.isclose(_norm, 1.0, rtol=1e-4):
                 probabilities = {k: v/_norm for k, v in probabilities.items()}
             else:
                 raise ValueError(f"Probabilities do not add up to 1: {probabilities}")
--- a/azure-quantum/azure/quantum/qiskit/job.py
+++ b/azure-quantum/azure/quantum/qiskit/job.py
@@ -145,6 +145,12 @@ class AzureQuantumJob(JobV1):
 
     @staticmethod
     def _draw_random_sample(sampler_seed, probabilities, shots):
+        _norm = sum(probabilities.values())
+        if _norm != 1 and np.isclose(_norm, 1.0):
+            probabilities = {k: v/_norm for k, v in probabilities.items()}
+        else:
+            raise ValueError(f"Probabilities do not add up to 1: {probabilities}")
+
         rand = np.random.RandomState(sampler_seed)
         rand_values = rand.choice(list(probabilities.keys()), shots, p=list(probabilities.values()))
         return dict(zip(*np.unique(rand_values, return_counts=True)))
--- a/azure-quantum/azure/quantum/qiskit/job.py
+++ b/azure-quantum/azure/quantum/qiskit/job.py
@@ -146,10 +146,11 @@ class AzureQuantumJob(JobV1):
     @staticmethod
     def _draw_random_sample(sampler_seed, probabilities, shots):
         _norm = sum(probabilities.values())
-        if _norm != 1 and np.isclose(_norm, 1.0):
-            probabilities = {k: v/_norm for k, v in probabilities.items()}
-        else:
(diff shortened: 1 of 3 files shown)


azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -142,6 +142,13 @@ def _format_results(self, sampler_seed=None):
 
     @staticmethod
     def _draw_random_sample(sampler_seed, probabilities, shots):
+        _norm = sum(probabilities.values())
+        if _norm != 1:
```
ricardo-espinoza, line 146: Question, how common is that we have an exact value of 1 here to check for equality explicitly instead of letting the logic below with numpy.isclose() taka care of it even if we end up dividing by 1 in rare ocassions?

ricardo-espinoza (approved):

azure-quantum/azure/quantum/qiskit/job.py
guenp: I found that it's pretty rare for the sum to not be equal to 1 hence the if statement.

anpaz (approved):

anpaz:
Should we normalize this in other places too? Like when using pass-through or Cirq?

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -142,6 +142,13 @@ def _format_results(self, sampler_seed=None):
 
     @staticmethod
     def _draw_random_sample(sampler_seed, probabilities, shots):
+        _norm = sum(probabilities.values())
```
anpaz, line 145: Should we normalize this in other places too? Like when using pass-through or Cirq?

guenp: Yes, thanks for pointing that out!

raghavthind2005 (approved):

azure-quantum/azure/quantum/qiskit/job.py
guenp: For cirq, we are using a package `cirq-ionq`. It looks like it does not normalize the probabilities. It would be worth looking into: https://github.com/quantumlib/Cirq/blob/05418b1996129298321435e9a566487968fcf126/cirq-ionq/cirq_ionq/job.py#L230
I suggest following up in that repo though and not in `azure-quantum`. What do you think?

PR #307, opened 2022-05-01
feat: Add `substitutions` parameter to `target.rigetti.InputParams`

dbanty:
This parameter allows users to submit multiple sets of parameters with the same circuit as a single job—leading to much higher performance.

### DIFF
--- a/azure-quantum/azure/quantum/target/rigetti/target.py
+++ b/azure-quantum/azure/quantum/target/rigetti/target.py
@@ -72,7 +72,8 @@ class InputParams:
         InputParams(
             substitutions={
                 "theta": [
-                    [0.0, np.pi], [np.pi, 0.0]
+                    [0.0, np.pi],
+                    [np.pi, 0.0],
                 ],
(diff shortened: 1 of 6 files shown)


azure-quantum/azure/quantum/target/rigetti/target.py
```diff
@@ -47,6 +47,54 @@ class InputParams:
...
+    
+    might be run with 
+    
+    .. highlight:: python
+    .. code-block::
+    
+        InputParams(
+            substitutions={
+                "theta": [
+                    [0.0, np.pi], [np.pi, 0.0]
```
anpaz, line 75: nit: just to keep it consistent with below.
```suggestion
                    [0.0, np.pi],
                    [np.pi, 0.0],
```

anpaz (approved):
lgtm

anpaz:
@dbanty. You'll need to pull from main to pick a fix for a unittest that started failing when the recording became stale.

PR #310, opened 2022-05-04
Fixing job filter test

anpaz:
In #309 I disabled the job filter tests when run with recordings as they would fail if the job creation time became stale, but now when they run live, the test fails as it was doing the wrong checks.
This fixes the test when run live.

### DIFF
--- a/azure-quantum/tests/unit/recordings/test_job_submit_microsoft_simulated_annealing.yaml
+++ b/azure-quantum/tests/unit/recordings/test_job_submit_microsoft_simulated_annealing.yaml
@@ -5,7 +5,7 @@ interactions:
       Accept:
       - application/json
       Accept-Encoding:
-      - gzip, deflate, br
+      - gzip, deflate
       Connection:
       - keep-alive
       Content-Length:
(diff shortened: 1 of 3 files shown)


azure-quantum/tests/unit/recordings/test_job_submit_microsoft_simulated_annealing.yaml
```diff
@@ -5,15 +5,15 @@ interactions:
...
+      - gzip, deflate
       Connection:
       - keep-alive
       Content-Length:
       - '287'
       Content-Type:
       - application/x-www-form-urlencoded
       User-Agent:
-      - azsdk-python-identity/1.7.1 Python/3.7.12 (Windows-10-10.0.19041-SP0)
+      - azsdk-python-identity/1.10.0 Python/3.7.8 (Windows-10-10.0.19041-SP0)
```
ricardo-espinoza, line 16: It seems that the recordings were generated this time using Python 3.7.8 which is a downgrade from 3.7.10 in the previous ones? Is this intentional? I don't think we need to regenerate them for this but wanted to double check.

ricardo-espinoza (approved):

azure-quantum/tests/unit/recordings/test_job_submit_microsoft_simulated_annealing.yaml
anpaz: That's the version I have installed. I should probably update my environement, lol.

PR #311, opened 2022-05-05
Fixing merge to main that causes tests to fail.

anpaz:


cgranade (approved):

ricardo-espinoza:
LGTM, just checked that none of the callers assume this to be static.

ricardo-espinoza (approved):

PR #314, opened 2022-05-11
Azure quantum aio (async) fixes

anpaz:
Found a couple of places where the async API was either broken or incorrect, in particular:
* StreamingProblem was not working correctly. The sync version runs the upload in a different thread; however since the code in aio is async by definition it can be easily simplified
* UploadAttachment: It had a bug in the code 
* QIO job submission: some changes on the sync version were not propagated causing problems on how the payload is uploaded and encoded.

Also, added a lot of missing `close()` calls and fixed tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -402,15 +402,14 @@ temp/
 
 # azure-quantum
 version.py
-!recordings/*.yaml
 azure-quantum/azure/quantum/version.py
 azure-quantum/dist
 azure-quantum/temp
 azure-quantum/build
 !azure-quantum/tests/unit/recordings/*.yaml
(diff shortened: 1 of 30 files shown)


ricardo-espinoza:
LGTM, focused mostly on commits [37203446](https://github.com/microsoft/qdk-python/pull/314/commits/3720344698e2923bef2b88c598a2005b09869938) and [efdd8730](https://github.com/microsoft/qdk-python/pull/314/commits/efdd87307637e7f4f68ff79ecec091a4fd539741).

ricardo-espinoza (approved):

PR #315, opened 2022-05-11
Update lookup of platform for Simulated Annealing to work correctly with experimental targets.

Anatoliy-Litvinenko:
Update parse for hardware target by .endswith(“cpu”) or .endswith(“fpga”) to correctly work with experimental targets.


azure-quantum/azure/quantum/target/microsoft/qio/simulated_annealing.py
```diff
@@ -109,7 +109,7 @@ def from_target_status(
         :return: Target instance
         :rtype: Target
         """
-        if status.id.endswith("cpu"):
+        if status.id.endswith("cpu") or status.id.endswith("cpu.experimental"):
             platform = HardwarePlatform.CPU
         elif status.id.endswith("fpga"):
```
xinyi-joffre, line 114: Do we expect for there to be an fpga.experimental in the future as well? Maybe we should add it just in case?

xinyi-joffre (approved):

azure-quantum/azure/quantum/target/microsoft/qio/simulated_annealing.py
Anatoliy-Litvinenko: Currently ...fpga.experimental is not implemented and will raise unknown target error, which is correct behavior as of today.

xinyi-joffre: Sounds good!

Issue #220, opened 2022-01-20
Latest version of azure-core (1.21.2) does not work with sanitized recordings


guenp:
See this commit: https://github.com/microsoft/qdk-python/pull/218/commits/1b9d8d339c306a29cda05c3729a1f95876dee66b

Workaround: https://github.com/microsoft/qdk-python/pull/218/commits/b96eaf472fdbc3cd9eaffd3daf7d32acdad95549

azure-core 1.21.2 checks the http response if the payload size matches the size in the header. since we have a logic in VCR to sanitize the recordings, the size of the payload is not matching

Reported by @vxfield

PR #318, opened 2022-05-12
Fix azure-storage-blob version and max version for protobuf

anpaz:
The new version of azure-storage-blob (12.12.0) requires azure-core>=1.23.1, but becuase #220 we can't pick it up.
Changing the conda environments to make sure the tests run successfully.


anpaz:
Turns out that for cirq we were also pulling a new breaking version (4.0) of protobuf which made other tests to fail as well.

ricardo-espinoza (approved):

PR #332, opened 2022-06-13
remove experimental targets and associated tests

frtibble:
removes "experimental" target from QDK, as experimental solvers have been switched to default


frtibble:
@Anatoliy-Litvinenko - tagging as an FYI on these changes

PR #333, opened 2022-06-13
add Tabu legacy tests

frtibble:
adds test for Tabu "legacy" targets


frtibble:
@Anatoliy-Litvinenko - FYI on this change

PR #335, opened 2022-06-13
Update attempt for qiskit test for removing experimental target

Anatoliy-Litvinenko:


anpaz (commented):
Instead of manually updating recordings, please run the tests in live mode and.

anpaz (changes requested):
Instead of manually updating recordings, please run the tests in live mode and.

PR #337, opened 2022-06-14
Update QMC solvers test calls to have trotter > 1

Anatoliy-Litvinenko:
QMC method cannot be called QMC if trotter is <=1, so all test calls should use bigger then 1 trotter.

### DIFF
--- a/azure-quantum/tests/unit/aio/test_aio_job.py
+++ b/azure-quantum/tests/unit/aio/test_aio_job.py
@@ -29,7 +29,7 @@ SOLVER_TYPES = [
     functools.partial(microsoft.SimulatedAnnealing, beta_start=0),
     functools.partial(microsoft.ParallelTempering, sweeps=100),
     functools.partial(microsoft.Tabu, sweeps=100),
-    functools.partial(microsoft.QuantumMonteCarlo, trotter_number=1),
+    functools.partial(microsoft.QuantumMonteCarlo, trotter_number=4),
     functools.partial(microsoft.PopulationAnnealing, sweeps=200),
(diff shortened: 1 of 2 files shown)


azure-quantum/tests/unit/aio/test_aio_job.py
```diff
@@ -29,7 +29,7 @@
     functools.partial(microsoft.SimulatedAnnealing, beta_start=0),
     functools.partial(microsoft.ParallelTempering, sweeps=100),
     functools.partial(microsoft.Tabu, sweeps=100),
-    functools.partial(microsoft.QuantumMonteCarlo, trotter_number=1),
+    functools.partial(microsoft.QuantumMonteCarlo, trotter_number=4),
```
xinyi-joffre, line 32: Do you know if we have to update the test recordings as well (cc: @ricardo-espinoza)? I think you can regenerate if you delete the recordings folder and then re-run tests, but I'm not super familiar with this part of code-base.

xinyi-joffre (approved):

azure-quantum/tests/unit/aio/test_aio_job.py
Anatoliy-Litvinenko: There 6 other methods which had similar changes like this one (Switch from QIO to QIOTE, like test_job_submit_microsoft_parallel_tempering). If changes to recording would be needed, they would fail.

frtibble: @xinyi-joffre we had a similar question for another PR too, I believe it is fine for this one - but I will start a thread to discuss

PR #342, opened 2022-06-22
fix initial config formatting in StreamingProblem

frtibble:


vxfield:
This is old/stale, so I'm closing it.

PR #349, opened 2022-06-30
Update to Cirq v0.15.0

ricardo-espinoza:


azure-quantum/environment-cirq-beta.yml
```diff
@@ -17,8 +17,8 @@ dependencies:
     - azure-core==1.19.1 # see https://github.com/microsoft/qdk-python/issues/220 
     - azure-devtools>=1.2.0
     - azure-storage-blob==12.11.0 # see https://github.com/microsoft/qdk-python/issues/220
-    - --pre cirq-core>=0.13.1
-    - cirq-ionq>=0.13.1
+    - --pre cirq-core>=0.15.0
```
anpaz, line 20: the point of this environment is to test Cirq with their beta version, so this should be pointing to 0.16 (it should point to 0.15 earlier, which would have allowed us to fix the problem before).

azure-quantum/setup.py
```diff
@@ -86,8 +86,8 @@
             "qiskit-qir>=0.1.0b12"
         ],
         "cirq": [
-            "cirq-core==0.14.1",
-            "cirq-ionq==0.14.1",
+            "cirq-core==0.15.0",
```
anpaz, line 89: I think this should stay on 0.14.1, as that's the minimal version we know it works with.

azure-quantum/environment.yml
```diff
@@ -17,8 +17,8 @@ dependencies:
     - azure-core==1.19.1 # see https://github.com/microsoft/qdk-python/issues/220
     - azure-devtools>=1.2.0
     - azure-storage-blob==12.11.0 # see https://github.com/microsoft/qdk-python/issues/220
-    - cirq-core>=0.13.1
-    - cirq-ionq>=0.13.1
+    - cirq-core>=0.15.0
```
anpaz, line 20: I think this one in 0.15.0 is the correct, as that is the version we want to test with

azure-quantum/environment-cirq-beta.yml
ricardo-espinoza: Interesting that we had both in 0.13.1.
All right, let me update this one.

azure-quantum/setup.py
ricardo-espinoza: @anpaz, tests are passing with 0.15.0, what validations do we need to go forward with the update from 0.14.1 to 0.15.0?
Since we're early in the release cycle, it might be a good time to update since we have a month to catch and fix possible issues with the upgrade.

ricardo-espinoza: Talked about this offline. :)

anpaz (approved):

PR #350, opened 2022-07-01
Update QDK Python CI tests to validate the next Cirq version

ricardo-espinoza:


ricardo-espinoza:
This has been abandoned since the work was done in another PR.

PR #362, opened 2022-08-03
Add support for new target IonQ Aria QPU

ricardo-espinoza:


azure-quantum/azure/quantum/target/ionq.py
```diff
@@ -12,6 +12,24 @@
 from azure.quantum.workspace import Workspace
 from azure.quantum._client.models import CostEstimate, UsageEvent
 
+COST_1QUBIT_GATE_MAP = {
```
anpaz, line 15: nitpick: I would group these by target so every time we add a new target we add one structure instead of modifying 3.

azure-quantum/azure/quantum/target/ionq.py
```diff
@@ -94,19 +113,16 @@ def estimate_cost(
...
+        price_2q: float = None,
+        min_price: float = None
     ) -> CostEstimate:
-        """Estimate the cost of submittng a circuit to IonQ targets.
+        """Estimate the cost of submitting a circuit to IonQ targets.
         Optionally, you can provide the number of gate and measurement operations
         manually.
         The actual price charged by the provider may differ from this calculation.
         
         Specify pricing details for your area to get most accurate results.
```
anpaz, line 125: I would add a comment about *the default price is taken from...*

azure-quantum/tests/unit/test_qiskit.py
```diff
@@ -117,6 +117,16 @@ def test_plugins_estimate_cost_qiskit_ionq(self):
         cost = backend.estimate_cost(circuit, shots=100e3)
         assert np.round(cost.estimated_total) == 66.0
 
+        ## The following two tests are skipped until we can use a workspace
```
anpaz, line 120: optionally, we can add a new tag `ionq.aria` that is part of a custom capability so the tests are only run when this target is present in the workspace.

anpaz (approved):

azure-quantum/azure/quantum/target/ionq.py
ricardo-espinoza: Sounds good. Adding.

azure-quantum/tests/unit/test_qiskit.py
ricardo-espinoza: Makes sense, but if we're planning to add Aria to the default test SKU, probably having a target specific tag will only have a short-lived effect, and that's not a common approach for others. If by next month Aria is not available, they I'll revisit this and add the tag, otherwise I'll just enable for `pytest.mark.ionq` in general.

PR #366, opened 2022-08-13
Qiskit expects literal bitstring results

anpaz:
QIR returns its output as an array of bits, as in: `[0, 1, 0]`
We use these as the keys for the counts and probabilities dictionaries in the Results object, however, Qiskit normally returns the keys in the values to be a literal bitstring and certain libraries (like VQE) assumes this always be the case, so if the value is differently it will fail.
This change removes the superfluos characters and makes sure the key is just a bitstring.

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/job.py
+++ b/azure-quantum/azure/quantum/qiskit/job.py
@@ -213,7 +213,8 @@ class AzureQuantumJob(JobV1):
         if (len(histogram) % 2) == 0:
             items = range(0, len(histogram), 2)
             for i in items:
-                bitstring = re.sub("[^01]", "", histogram[i])  # Qiskit expects a literal bitstring, QIR returns it as an array, remove superfluos characters.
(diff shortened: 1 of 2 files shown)


azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -212,7 +213,7 @@ def _format_microsoft_results(self, sampler_seed=None):
         if (len(histogram) % 2) == 0:
             items = range(0, len(histogram), 2)
             for i in items:
-                bitstring = histogram[i]
+                bitstring = re.sub("[^01]", "", histogram[i])  # Qiskit expects a literal bitstring, QIR returns it as an array, remove superfluos characters.
```
cgranade, line 216:
```suggestion
                bitstring = re.sub("[^01]", "", histogram[i])  # Qiskit expects a literal bitstring, QIR returns it as an array, remove superfluous characters.

```

cgranade (approved):
LGTM, but slightly concerned about what happens if Qiskit changes their data model here; would it be possible to add some integration tests in addition to the unit tests? Thanks!

PR #369, opened 2022-08-25
remove platform param from solver 0

sanjgupt:


vxfield:
This is old/stale, so I'm closing it.

PR #376, opened 2022-09-14
Update version of cirq, azure-storage-blob and azure-core packages

ricardo-espinoza:


vxfield:
@ricardo-espinoza, please see this issue that is related to the azure-core package version and the test failures dues to `different payload size expected`
#220

ricardo-espinoza:
> @ricardo-espinoza, please see issue #220 that is related to the azure-core package version and the test failures dues to `different payload size expected`

I'm aware of that issue, and that's why this is still a draft. I'll reach out to you offline to discuss options. Unfortunately, we need to move on from version 12.11.0, and find a way to do so without regressing on tests.

anpaz (approved):
Looks great!

PR #377, opened 2022-09-14
azure to qiskit reformat

### LINKED ISSUE #361: endianness of classical registers
If I execute a circuit that looks like this in Azure Quantum:

```
2022-08-01 09:15:28,499 - INFO - __main__ - ================== Running circuit endian0cr3:
2022-08-01 09:15:28,499 - INFO - __main__ -
      ┌───┐┌─┐
q3_0: ┤ X ├┤M├
      └───┘└╥┘
q3_1: ──────╫─
            ║
q3_2: ──────╫─
            ║
c3: 3/══════╩═
            0

c4: 3/════════

c5: 3/════════
```

created like this:
```python
        pos = 0
        qr = QuantumRegister(3)
        cr = [ClassicalRegister(3) for _ in range(3)]
        qc = QuantumCircuit(qr, *cr, name=f"endian{pos}cr3")
        qc.x(pos)
        qc.measure(qr[pos], cr[pos][pos])
```

Which I then execute against both `qci.simulator` and `qasm_simulator` to compare.

When running on `azure-quantum[qiskit]` version `0.25.222597`, I get the result:
```
{'[0, 0, 1], [0, 0, 0], [0, 0, 0]': 100}
```
while on `qasm_simulator` I get:
```
{'000 000 001': 10000}
```

That is, the ordering within each classical register matches, but the ordering of the classical registers themselves are reversed.

Is this a bug or is this just a difference in conventions?

Also, as an aside, it's worth pointing out that I have my own code to convert conventions from "tuple of lists" to a "space separated bits". In this code, I simply reverse the order I traverse tuples in order to get results I can compare with `hellinger_fidelity`.

### LINKED ISSUE #375: shouldn't remove the register separators from the histogram keys
https://github.com/microsoft/qdk-python/blob/a3230d8356120f96f563d49affd9cc835c2cf6b8/azure-quantum/azure/quantum/qiskit/job.py#L216

this line, introduced in #366 does not account for multiple classical registers. Qiskit separates them with a space in the histogram key.

this is partially related to issue #361

qci-amos:
closes #375 
closes #361


azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -197,6 +198,23 @@ def _format_ionq_results(self, sampler_seed=None):
 
         return {"counts": counts, "probabilities": probabilities}
 
+    @staticmethod
+    def _azure_to_qiskit(obj):
```
anpaz, line 202: I would name this something like `_to_qiskit_bitstring`, as this converts the result key from the QIR format to one supported by Qiskit.

qci-amos: what do you think of my counter proposal?

azure-quantum/tests/unit/test_qiskit.py
```diff
@@ -659,3 +680,36 @@ def test_qiskit_get_qci_qpu_targets(self):
...
+            assert qiskit_job._azure_job.details.metadata["num_qubits"] == '3'
+
+            # Make sure the job is completed before fetching the results
+            self._qiskit_wait_to_complete(qiskit_job, provider)
+
+            assert JobStatus.DONE == qiskit_job.status()
+            result = qiskit_job.result()
+            print(result)
+            assert sum(result.data()["counts"].values()) == shots
+            assert result.data()["counts"][expectation] == shots
```
anpaz, line 715:
```suggestion

            if JobStatus.DONE == qiskit_job.status():
                result = qiskit_job.result()
                print(result)
                assert sum(result.data()["counts"].values()) == shots
                assert result.data()["counts"][expectation] == shots
```
Instead of `assert`, could you please just check status and check results only if job succeeds.
When we run these tests live there might be some service problems, so we just report warnings if the job fails/timeouts. We only check results if the job actually succeeds

anpaz (changes requested):
Just a very small check at the end, but overall looks great!

anpaz (approved):
Looks great! Thanks!

PR #383, opened 2022-09-14
Fix #220 on main branch

vxfield:
This PR fixes #220 by [correctly setting the `content-length` HTTP header](https://github.com/microsoft/qdk-python/pull/383/commits/8459cabdd27f2ec0229d6cb15522f1d5d4a74ca2) when sanitizing our test recordings.
It also contains re-recorded unit tests with the correct `content-length` header.

PR #385, opened 2022-09-15
Address comments from #383

vxfield:


ricardo-espinoza:
Shouldn't this be ready for review already?

ricardo-espinoza:
LGTM, couldn't find any secrets in URLs.

PR #386, opened 2022-09-15
add noisy qci backend

qci-amos:


qci-amos:
can you review, @anpaz ? it's not letting me select reviewers.

fwiw, I verified that this works from a jupyter notebook.

anpaz (approved):

PR #396, opened 2022-10-13
Remove Honeywell provider

vxfield:
### Context
In November 2021, Honeywell became Quantinuum: Press release: [Quantinuum is a result of the combination of two global leaders in quantum computing: Honeywell Quantum Solutions and Cambridge Quantum](https://www.quantinuum.com/pressrelease/introducing-quantinuum)

In February/March 2022, Azure Quantum started supporting the Quantinuum provider and targets.

### This PR
This PR removes the Honewell provider, targets and tests.


ricardo-espinoza:
@vxfield , I think this will require an update to the recordings, and at the same time eliminating those from Honeywell tests.

vxfield:
@ricardo-espinoza, yes, work-in-progress :-)

vxfield:
@ricardo-espinoza, ready for final review now

azure-quantum/azure/quantum/qiskit/backends/quantinuum.py
```diff
@@ -52,18 +52,12 @@
 QUANTINUUM_PROVIDER_ID = "quantinuum"
 QUANTINUUM_PROVIDER_NAME = "Quantinuum"
 
-HONEYWELL_PROVIDER_ID = "honeywell"
-HONEYWELL_PROVIDER_NAME = "Honeywell"
-
 class QuantinuumBackend(AzureBackend):
-    """Base class for interfacing with a Quantinuum (formerly Quantinuum) backend in Azure Quantum"""
+    """Base class for interfacing with a Quantinuum (formerly Honeywell) backend in Azure Quantum"""
```
ricardo-espinoza, line 56: Ha! Thanks for catching that. In any case, we should remove the "(formerly Honeywell)" entirely here since it's gone from other parts of the code. No need to block the PR for this, so I'll signoff anyway.

ricardo-espinoza (approved):

azure-quantum/azure/quantum/qiskit/backends/quantinuum.py
vxfield: Ok. I'll let it there then and we can remove at the next PR.

PR #398, opened 2022-10-20
Updates to resource estimation output

msoeken:


fedonman (approved):

azure-quantum/azure/quantum/qiskit/results/resource_estimator.py
```diff
@@ -6,7 +6,63 @@
...
+        html = ""
+
+        md = markdown.Markdown(extensions=['mdx_math'])
+        for group in self.data()['reportData']['groups']:
+            html += f"""
+                <details {"open" if group['alwaysVisible'] else ""}>
+                    <summary style="display:list-item">
+                        <strong>{group['title']}</strong>
+                    </summary>
+                <table>"""
```
kuzminrobin, line 32: (Nit) For me the `<table>` indentation in line 32 (L32) is shown the same as the indentation for `<details` in L28. This makes an impression of a mismatch between the opening and closing tag.
Consider increasing the indentation for `<table>` (to the same as for `</summary>` in L31).

anpaz, line 32: If this is markup, is there a reason to use `<table>` elements instead of markup?
```
| h1 | h2 |
|----|----|
| v1 | v2 |
```

anpaz (changes requested):
These new requirements seem to be only for the qiskit integration; as such instead of adding them to requirements.txt they should go directly in setup.py, under `extras_require`
https://github.com/microsoft/qdk-python/blob/main/azure-quantum/setup.py#L84

azure-quantum/azure/quantum/qiskit/results/resource_estimator.py
```diff
@@ -6,7 +6,63 @@
...
+            for entry in group['entries']:
+                val = self.data()
+                for key in entry['path'].split("/"):
+                    val = val[key]
```
kuzminrobin, line 36: I don't know what L36 does, but it looks suspicious, especially in a loop (a variable's value is overwritten with a part of itself).

azure-quantum/azure/quantum/qiskit/results/resource_estimator.py
msoeken: The table is making use of HTML features that cannot be accessed through Markdown

azure-quantum/azure/quantum/qiskit/results/resource_estimator.py
msoeken: It's is recursively entering a Python dictionary based on a path, e..g.,

`D = {'a': {'b': {'c': 42}}}`

with

`path = "a/b/c"`

would evaluate to `42`.

where `val` would be `D`, then `{'b': {'c': 42}}`, then `{'c': 42}` and finally `42`.

azure-quantum/azure/quantum/qiskit/results/resource_estimator.py
msoeken: Fixed this.

anpaz (approved):

PR #400, opened 2022-10-27
E2E tests for resource estimation

msoeken:


azure-quantum/pytest.ini
```diff
@@ -11,3 +11,4 @@ markers =
     quantinuum: mark a test as requiring access to quantinuum.
     rigetti: mark a test as requiring access to Rigetti targets.
     qci: mark a test as requiring access to QCI targets.
+    microsoft_qc: mark a test as requiring access to microsoft-qc targets.
```
xinyi-joffre, line 14: Is the underscore because dashes do not work for pytest marks?

xinyi-joffre (approved):

azure-quantum/pytest.ini
msoeken: Yes, exactly.

PR #401, opened 2022-10-28
Use basis_gates for qiskit circuit transpilation.

idavis:
This changes each QIR backend to use its own basis gates as the providers don't use a consistent set. It also runs a pass to remove barriers as `transpile` will not remove them despite what is supplied in the basis gates.

This PR will also update the version of qiskit-qir used which has `ccx`, `swap`, and `barrier` support.


msoeken (approved):

fedonman (approved):

idavis:
@anpaz @msoeken I made a change to the qiskit transpile to default the opt level to 0 before QIR generation. Just wanted to run it by you before merging.

cesarzc (approved):

PR #406, opened 2022-12-06
Submit QIR Job From Qiskit to IonQ

ScottCarda-MS:
Allows for the submission of QIR jobs from Qiskit python to IonQ targets.

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/backends/ionq.py
+++ b/azure-quantum/azure/quantum/qiskit/backends/ionq.py
@@ -53,9 +53,9 @@ class IonQBackend(AzureBackend):
 
     def _translate_input(self, circuit, data_format, input_params, to_qir_kwargs={}):
         """ Translates the input values to the format expected by the AzureBackend. """
-        if "targetCapability" in input_params and input_params["targetCapability"] == "BasicExecution":
+        if "targetCapability" in input_params:
             return super()._translate_input(circuit, "qir.v1", input_params, to_qir_kwargs)
-        elif data_format == "ionq.circuit.v1":
+        else:
             ionq_circ, _, _ = qiskit_circ_to_ionq_circ(circuit, gateset=self.gateset())
             input_data = {
                 "gateset": self.gateset(),
@@ -63,8 +63,6 @@ class IonQBackend(AzureBackend):
                 "circuit": ionq_circ,
             }
             return (IonQ._encode_input_data(input_data), data_format, input_params)
-        else:
-            return super()._translate_input(circuit, data_format, input_params, to_qir_kwargs)
 
     def gateset(self):
         return self._gateset
--- a/azure-quantum/azure/quantum/qiskit/backends/ionq.py
+++ b/azure-quantum/azure/quantum/qiskit/backends/ionq.py
@@ -31,7 +31,7 @@ class IonQBackend(AzureBackend):
 
     @classmethod
     def _default_options(cls):
-        return Options(shots=500)
+        return Options(shots=500, targetCapability="ionq")
 
     @classmethod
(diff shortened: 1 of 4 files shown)


fedonman (approved):

azure-quantum/azure/quantum/target/microsoft/qio/simulated_annealing.py
```diff
@@ -16,10 +16,10 @@
 
 
 class SimulatedAnnealing(Solver):
-    target_names = [
+    target_names = (
```
msoeken, line 19: Is this change intended? Just asking because the code is out of the scope for the PR tilte.

msoeken (approved):

azure-quantum/azure/quantum/target/microsoft/qio/simulated_annealing.py
ScottCarda-MS: It is not related. I was fixing a bug where a target_name was just a single element and not being treated as a list, and saw that this was the only target_names that was a list instead of a tuple. I see that the bug was independently fixed while I was working on this in the same way.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -53,7 +53,9 @@ def _prepare_job_metadata(self, circuit, **kwargs):
 
     def _translate_input(self, circuit, data_format, input_params, to_qir_kwargs={}):
         """ Translates the input values to the format expected by the AzureBackend. """
-        if data_format == "ionq.circuit.v1":
+        if "targetCapability" in input_params and input_params["targetCapability"] == "BasicExecution":
```
DmitryVasilevsky, line 56: So targetCapability can be "ionq" (line 34) or "BasicExecution"? Or these are different target capabilities?

ScottCarda-MS: Yes that's right. I didn't know what to use for the default target capability for ionq, but I saw that quantuniuum has default capability of "openqasm", so I went with "ionq" for the default for ionq. It isn't a capability that would be recognized, I don't imagine, but the point is for there to be a default option other than "BasicExecution".

If there is a more correct way to set this up, please let me know, because I just took a guess here.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -31,7 +31,7 @@ class IonQBackend(AzureBackend):
 
     @classmethod
     def _default_options(cls):
-        return Options(shots=500)
+        return Options(shots=500, targetCapability="ionq")
```
cesarzc, line 34: Is there a reason for specifying the `targetCapability` by default? I believe the previous behavior should still be the default.

cesarzc:
@anpaz, could you please take a look? I did an initial review but since you are more familiar with the code base it would be good if you could also review it.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
ScottCarda-MS: The behavior is the same by default. Only if someone sets targetCapability to "BasicExecution" is the qir translation triggered.

ScottCarda-MS: I've removed the default targetCapability for ionq since it was causing confusion.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
ScottCarda-MS: I've removed the default targetCapability.

PR #407, opened 2022-12-14
Added Default Target Capability to IonQ

ScottCarda-MS:
The IonQ target needs to have a default targetCapability so that targetCapability can propagate through to the translation.


swernli (approved):
LGTM!

PR #408, opened 2022-12-20
Fix link in Python package description.

anpaz:
The link to the contributing guidelines was broken on the pypi package, as it was a relative link to the CONTRIBUTING.md file. This updates the README to point to the full URL.

### DIFF
--- a/azure-quantum/README.md
+++ b/azure-quantum/README.md
@@ -79,7 +79,7 @@ result = job.get_results()
 
 ## Contributing ##
 
-For details on contributing to this repository, see the [contributing guide](https://github.com/microsoft/qdk-python/blob/main/CONTRIBUTING.md).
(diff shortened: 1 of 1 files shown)


azure-quantum/README.md
```diff
@@ -79,7 +79,7 @@ result = job.get_results()
 
 ## Contributing ##
 
-For details on contributing to this repository, see the [contributing guide](../CONTRIBUTING.md).
+For details on contributing to this package, see the [contributing guide](https://github.com/microsoft/qdk-python/blob/main/CONTRIBUTING.md).
```
viktorveis, line 82: I suggest updating the text so that it appropriate for both a repository on github and a package on PyPi.

anpaz: The README is only for the azure-quantum package, not for the entire repo, so I think it is ok to focus on just package.

viktorveis (approved):

PR #409, opened 2023-01-05
Enable Batch Job Submission for Qiskit

ScottCarda-MS:
Translates the qiskit job input to the `microsoft.quantum-results.v2` format before submission.
Translates the qiskit job output to the format expected by qiskit after job is processed by target.


idavis:
Replace by #420

PR #411, opened 2023-01-09
More details about the VCR testing recording framework

vxfield:
This PR adds more details about the VCR testing recording framework including:

1) The common error of "Cannot Overwrite Existing Cassette Exception" and how to fix it
2) A paragraph about recording sanitization


azure-quantum/tests/README.md
```diff
@@ -43,6 +44,35 @@ For example, when a job is submitted and we want to fetch the job status, the HT
...
+
+But it does not handle the case when the recording file is there but a specific HTTP request/response can’t be found in the existing recording, which gives the error you are seeing.
+
+This error could also be caused if the recorded files are manually updated and do not really match the requests that the SDK will actually request.
+
+#### Potential solutions ####
+1) One way to remove the error is to delete the existing recording file and let it do all the live calls, and create a new recording file that contains all the requests/responses that the tests need. Then, next time, you should be able to simple playback the recordings with no errors.
+
+2) If the error still persist after trying (1), then probably there is something unique in the URL or HTTP headers of the HTTP request that changes at every time you run the tests. In this case, we need to either make that thing constant in the tests, or, if they are genuinely unique, we need to replace that unique value in the request recording pipeline such that, at least in the recording file, it will be unique.
```
ScottCarda-MS, line 65:
```suggestion
2) If the error still persist after trying (1), then probably there is something unique in the URL or HTTP headers of the HTTP request that changes every time you run the tests. In this case, we need to either make that thing constant in the tests, or if they are genuinely unique, we need to replace that unique value in the request recording pipeline such that, at least in the recording file, it will be unique.
```

azure-quantum/tests/README.md
```diff
@@ -43,6 +44,35 @@ For example, when a job is submitted and we want to fetch the job status, the HT
...
+
+The environment variables to control whether to playback records or run live and re-create them is just a little logic to decide if we want to delete the recording files first before running the tests (which will force a call to the live APIs and re-record the calls).
+
+But it does not handle the case when the recording file is there but a specific HTTP request/response can’t be found in the existing recording, which gives the error you are seeing.
+
+This error could also be caused if the recorded files are manually updated and do not really match the requests that the SDK will actually request.
+
+#### Potential solutions ####
+1) One way to remove the error is to delete the existing recording file and let it do all the live calls, and create a new recording file that contains all the requests/responses that the tests need. Then, next time, you should be able to simple playback the recordings with no errors.
```
ScottCarda-MS, line 63:
```suggestion
1) One way to remove the error is to delete the existing recording file and let it do all the live calls and create a new recording file that contains all the requests/responses that the tests need. After that, you should be able to simple playback the recordings with no errors.
```

ScottCarda-MS (approved):
Added a suggestion to be more direct in the Cause section. Thanks for adding this!

PR #413, opened 2023-01-24
Tweak ionq cost.

anpaz:
Updating IonQ estimated cost of 1 qubit gate .


ffrachon (commented):
Verified with IonQ that the Aria pricing is indeed $0.0002205 for 1QGS.

xinyi-joffre (approved):

PR #417, opened 2023-02-08
[DRAFT] Python SDK support for Sessions

vxfield:
# Work in progress

To-do:
- [x] Regenerate the SDK REST client from the L2 swagger
- [X] Expose the session methods in the target for regular Azure Quantum Targets
- [ ] Add tests to session methods (regular Azure Quantum Targets)
- [X] Expose the session methods in the target for Qiskit Targets
- [ ] Add tests to session methods (Qiskit)
- [X] Expose the session methods in the target for Cirq Targets
- [ ] Add tests to session methods (Cirq)
- [ ] Add session documentation/examples for regular AQ, Qiskit and Cirq


vxfield:
@ashwinmayya, thank you for an early look in this PR.
It's still a draft and I'm not done coding.
Still, every early feedback is highly appreciated.

For now, please disregard the `tests\unit` folder.
It's a very early draft and I'll need to refactor a few things in the testing framework, to allow for multiple job submission testing/mocking.

Also, you don't need to worry to much about the `_client` folder, which is completely code generated from the Swagger.

azure-quantum/azure/quantum/target/target.py
```diff
@@ -2,19 +2,21 @@
...
 from azure.quantum.job.job import Job
+from azure.quantum.job.session import Session, SessionHost
 from azure.quantum.job.base_job import ContentType
+
 if TYPE_CHECKING:
     from azure.quantum import Workspace
 
-
-class Target:
+class Target(ABC, SessionHost):
```
ashwinmayya, line 19: What is this ABC?

azure-quantum/azure/quantum/job/session.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @abstractmethod
+    def _get_azure_target_id(self) -> str:
+        raise NotImplementedError
+
+    def start_session(
+        self,
+        session_details: Optional[SessionDetails] = None,
+        session_id: Optional[str] = None,
+        session_name: Optional[str] = None,
```
ashwinmayya, line 150: Just id and name

azure-quantum/azure/quantum/job/session.py
```diff
@@ -0,0 +1,165 @@
...
+        job_failure_policy: Union[str, SessionJobFailurePolicy, None] = None,
+        **kwargs
+    ) -> Session:
+        if self.current_session:
```
ashwinmayya, line 154: One thing that we could do is to check if current_session has been ended. If it has, we can create a new session. This is nice because we could have customer's wanting to do which seems like a reasonable expectation-

backend = provider.get_backend("...")
with backend.start_session() as session:
    backend.run(circuit1)
    backend.run(circuit2)

// Do some classical logic

with backend.start_session() as session:
    backend.run(circuit3)
    backend.run(circuit4)

azure-quantum/azure/quantum/target/target.py
vxfield: When I was learning Python I had exactly the same question :-)
https://docs.python.org/3/library/abc.html

azure-quantum/azure/quantum/job/session.py
vxfield: Instead of persisting the current session in the target, what do you think of not persisting it at all?

The alternative would be:
```python
with backend.start_session() as session:
     session.run(circuit3)
     session.run(circuit4)
```

That would reduce the complexity of managing the session state for a given target.

We can make it in a way that the session object returned by `target.start_session()` inherits from and behaves as the target object, with the difference that there is a session associated with it.

ashwinmayya: Since customers are used to doing backend.run() today and one of the top goals we have had is to require minimum code changes to existing code, I think session.run is not a feasible option.

azure-quantum/azure/quantum/job/session.py
```diff
@@ -0,0 +1,165 @@
...
+                           Cirq: a new target instance can be obtained with `service.get_target("provider_id.target_name")`.
+                           """
+        super().__init__(self.message)
+
+
+class SessionHost(Protocol):
+    _current_session: Optional[Session] = None
+
+    @property
+    def current_session(self) -> Optional[Session]:
```
ashwinmayya, line 128: The names "current_session" and "get_current_session_id" may give the false impression that the session is still open, when it is not necessarily true.

Consider-
with backend.start_session() as session:
  backend.run()

print(backend.current_session_id)
 
I think it may be better to replace the term "current" with "latest". I think my comment below about checking if the latest session is open or not when handling a new start_session still applies.

azure-quantum/azure/quantum/job/session.py
vxfield: Makes sense to reduce customer friction. I'll leave as-is.

azure-quantum/azure/quantum/job/session.py
vxfield: I changed it to "latest" session.
And, as discussed before, when there is already an open latest_session in the target, and the user opens a new sesssion, instead of throwing an error the logic will automatically close the previous latest_session and open a new one.

azure-quantum/azure/quantum/job/session.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @property
+    def target(self) -> "Target":
+        return self._target
+
+    def start(self) -> "Session":
+        self.workspace.start_session(self)
+        return self
+
+    def end(self) -> "Session":
```
ashwinmayya: References to "end" and "end_session" need to be updated to "close" and "close_session" everywhere

vxfield: done

PR #419, opened 2023-02-14
RE batching

msoeken:
Implements batching for resource estimation.

* [x] Resource estimation table for batching jobs (with indexing into single items)
* [x] Summary plot for a resource estimation batching result
* [x] E2E test for job with multiple items (required changes to `AccessTokenReplacer`)

Plot preview (click on image to see ticks in dark mode):

image

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/job.py
+++ b/azure-quantum/azure/quantum/qiskit/job.py
@@ -18,7 +18,7 @@ import ast
 import json
 import re
 from azure.quantum import Job
-from azure.quantum.qiskit.results.resource_estimator import ResourceEstimatorResult
+from azure.quantum.qiskit.results.resource_estimator import ResourceEstimatorResult, ResourceEstimatorBatchResult
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -95,7 +95,8 @@ def result(self, timeout=None, sampler_seed=None):
 
         result_type = Result
         if self._azure_job.details.output_data_format == RESOURCE_ESTIMATOR_OUTPUT_DATA_FORMAT:
-            result_type = ResourceEstimatorResult
+            is_simple_job = not isinstance(results['data'], list)
```
xinyi-joffre, line 98: Do you know if data is guaranteed to be on the result object? Will it still come back as empty array if job is not successful? What if it fails validation at data plane before reaching resource estimator?

xinyi-joffre (approved):

azure-quantum/azure/quantum/qiskit/job.py
msoeken: Yes, `data` is guaranteed to be on the `result` object and is an empty object in case of an error. However, I added an additional check to only check for batching jobs on successful jobs.

Issue #410, opened 2023-01-09
Support `azure-quantum[qiskit]` installation on macOS M1


jond01:
Installing the `azure-quantum` package with the `qiskit` extras on macOS with ARM processors (e.g. M1) is not supported currently, because of this fixed dependency:
https://github.com/microsoft/qdk-python/blob/bc4ae3a5bb1a84cfcb7f68c3e3e2f1d7f38f69f5/azure-quantum/setup.py#L86
So the dependency path looks like:
```
azure-quantum 0.27.244707 Python client for Azure Quantum
├── qiskit-qir >=0.2,<0.3
│   ├── pyqir-generator 0.4.2a1 
```

However, the latest version of `pyqir` supports ARM processors, see https://github.com/microsoft/qiskit-qir/issues/34 (which has to be addressed first).

jond01:
Ok, it looks like v0.27.238334 solved the issue:
https://github.com/microsoft/qdk-python/blob/975d08bf7aea9bae7778800a6ae0f5e00fc74cd2/azure-quantum/setup.py#L86

This is still relevant on the main branch, though

idavis:
@jond01 This will be handled in #420 which updates to PyQIR 0.8 and qiskit-qir 0.3.1

PR #420, opened 2023-02-26
Adding unique backends for QIR and pass through.

idavis:
# Overview
- Adding unique backends for QIR and pass through.
- Backend filtering now works.
- TargetFactory no longer does double duty for Backend creation.
- Support for output results format v2

# Questions
 - The signature for `BackendV1.run` is `def run(self, run_input, **options)` in the Qiskit API. Our current implementation uses `def run(self, circuit, **kwargs)` which is focused on a single circuit while other values can be passed. This PR updates the def of run to match the original API. This will effect anyone that tries to explicitly specify the circuit param as seen in our tests. AFAICT we don't use that syntax in our samples. Are we ok with this change?

# Examples
## Backend filtering
Filtering of backends from the provider now works according to the Qisksit backend definition. In addition to the backend name, `**kwargs` can now be passed that will return backends that match all criteria specified that are both available locally via implementations and available in the workspace:
```python
for backend in provider.backends(provider_id="quantinuum", n_qubits=20):
    print("- " + backend.name())
```
```shell
- quantinuum.hqs-lt-s1-apival
- quantinuum.hqs-lt-s1-apival
- quantinuum.sim.h1-1sc
- quantinuum.sim.h1-1sc
...
```

This looks like duplicates, however they are distinct backends for QIR and pass through. 

## Getting a specific backend
Calling `backends` will always return a list of all available backends.

```python
for backend in provider.backends("ionq.simulator"):
    print("- " + backend.name() + ", " + backend.configuration().azure["input_data_format"])
```
```shell
- ionq.simulator, ionq.circuit.v1
- ionq.simulator, qir.v1
```

This works normally for `provider.get_backend` as it always expects to return a single backend, and will throw an error if 0 or > 1 are returned.  To handle multiple backends with the same target name, one implementation is marked as default. If the filtered backends all have the same name, it applies an additional filter to include only the default backend for that name. As providers move to QIR as the default, we can change this configuration to switch which backend is returned.

```python
print(backend.name() + ", " + backend.configuration().azure["input_data_format"])
```
```log
ionq.simulator, ionq.circuit.v1
```
```python
backend = provider.get_backend("ionq.simulator", input_data_format="qir.v1")
print(backend.name() + ", " + backend.configuration().azure["input_data_format"])
```
```log
ionq.simulator, qir.v1
```

### DIFF
--- a/azure-quantum/azure/quantum/qiskit/backends/microsoft.py
+++ b/azure-quantum/azure/quantum/qiskit/backends/microsoft.py
@@ -3,8 +3,9 @@
 # Licensed under the MIT License.
 ##
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Dict, List
 from azure.quantum.version import __version__
+from qiskit import QuantumCircuit
 
 from .backend import AzureQirBackend
 
@@ -30,25 +31,25 @@ QIR_BASIS_GATES = [
     "x",
     "y",
     "z",
-    "id"
+    "id",
 ]
 
 if TYPE_CHECKING:
     from azure.quantum.qiskit import AzureQuantumProvider
 
 import logging
+
 logger = logging.getLogger(__name__)
 
-__all__ = [
-    "MicrosoftBackend", "MicrosoftResourceEstimationBackend"
-]
+__all__ = ["MicrosoftBackend", "MicrosoftResourceEstimationBackend"]
+
 
 class MicrosoftBackend(AzureQirBackend):
     """Base class for interfacing with a Microsoft backend in Azure Quantum"""
 
     @classmethod
     def _default_options(cls):
-        return Options(entryPoint="main", arguments=[], targetCapability="AdaptiveExecution")
+        return Options(targetCapability="AdaptiveExecution")
 
     def _azure_config(self):
         config = super()._azure_config()
@@ -56,10 +57,11 @@ class MicrosoftBackend(AzureQirBackend):
             {
                 "provider_id": "microsoft-qc",
                 "output_data_format": "microsoft.resource-estimates.v1",
-                "to_qir_kwargs": {"record_output": False}
+                "to_qir_kwargs": {"record_output": False},
             }
         )
         return config
+
 
 class MicrosoftResourceEstimationBackend(MicrosoftBackend):
     """Backend class for interfacing with the resource estimator target"""
@@ -73,9 +75,9 @@ class MicrosoftResourceEstimationBackend(MicrosoftBackend):
             arguments=[],
             targetCapability="AdaptiveExecution",
             errorBudget=1e-3,
-            qubitParams={"name":"qubit_gate_ns_e3"},
-            qecScheme={"name":"surface_code"},
-            items=None
+            qubitParams={"name": "qubit_gate_ns_e3"},
+            qecScheme={"name": "surface_code"},
+            items=None,
         )
 
     def __init__(self, name: str, provider: "AzureQuantumProvider", **kwargs):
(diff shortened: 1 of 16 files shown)


vxfield:
> Question: The signature for BackendV1.run is def run(self, run_input, **options) in the Qiskit API. Our current implementation uses def run(self, circuit, **kwargs) which is focused on a single circuit while other values can be passed. This PR updates the def of run to match the original API. This will effect anyone that tries to explicitly specify the circuit param as seen in our tests. AFAICT we don't use that syntax in our samples. Are we ok with this change?

I think it makes sense to use the qiskit signature as it's more general. 
Since it's a breaking change in our API, it may be better to support both the old and new signature and emit a deprecating warning for the old signature. Example [here](https://github.com/microsoft/qdk-python/blob/dd8854a9edec377ab2ff4ce2e141fae004a13423/azure-quantum/azure/quantum/target/oneqbit/solvers.py#L203)

vxfield:
Do you need to update or add unit tests to cover the previous behavior (to test for breaking changes) and to cover the new functionality?

idavis:
@vxfield Yes, I need to update the tests. I was waiting until this last refactoring as I'd have to rewrite them. That is the next thing on my list.

idavis:
@vxfield In order to use the deprecation scheme you mentioned we would need to make the `run_input` parameter have a default value of None or an empty list as it is a required positional parameter. Then we would check the `**options` for the existence of circuit and supply the warning. 
```python
    def run(self, run_input = [], **options):
        """Submits the given circuit to run on an Azure Quantum backend."""
        if "circuit" in options:
            warnings.warn(DeprecationWarning("The `ciruit` parameter has been deprecated and will be ignored. Please use `run_input` instead."))
            run_input = circuit
...
```
If the `run_input` isn't supplied the run can't continue. We can map circuit into run_input or try to fail. I'm not sure we can use the exact pattern you mentioned.

The QIR job run method has the same arguments, but supplies better type hints.  We'd have to make the default an empty list or it would change the type.

azure-quantum/azure/quantum/qiskit/provider.py
```diff
@@ -2,59 +2,91 @@
...
+
+        backends = self.backends(name=name, **kwargs)
+
+        if len(backends) > 1:
+            def all_same(iterable):
+                group_iter = groupby(iterable)
+                return next(group_iter, True) and not next(group_iter, False)
+
+            # If all backends have the same name, filter for default backend
+            if all_same(backend.name() for backend in backends):
```
ashwinmayya, line 66: Isn't it known at this point that all the backends in 'backends' have the same name? Given that we specify a name filter in line 58?

idavis: The `name` defaults to `None`: `def get_backend(self, name=None, **kwargs):`  - so we don't filter on name unless it is supplied. Line 58, `backends = self.backends(name=name, **kwargs)`, forwards the `name` or `None` if unset to the `self.backends`.

ashwinmayya (approved):

vxfield (approved):
Changes look good to me.

My only comment, as discussed offline, is about the unit tests and the tradeoffs between writing new provider/endpoint mocks vs. using a recorded/live "real" provider.

I think it's fine for this PR to introduce these new mock/dummy providers, but in the future we might want to rethink this to avoid the proliferation of new custom mocks as unit tests get introduced.
This is somewhat a technical debt that we have in our unit tests (not having some generic/flexible mock objects for tests that are 100% local and that do not need a service call), and we can tackle that in the future.

azure-quantum/azure/quantum/qiskit/backends/microsoft.py
```diff
@@ -90,22 +92,28 @@ def __init__(self, name: str, provider: "AzureQuantumProvider", **kwargs):
...
                 "memory": False,
-                "n_qubits": 0xffffffffffffffff, # NOTE: maximum 64-bit unsigned value
+                "n_qubits": 0xFFFFFFFFFFFFFFFF,  # NOTE: maximum 64-bit unsigned value
                 "conditional": True,
                 "max_shots": 1,
                 "max_experiments": 1,
                 "open_pulse": False,
-                "gates": [{"name": "TODO", "parameters": [], "qasm_def": "TODO"}], # NOTE: copied from other backends
-                "azure": self._azure_config()
+                "gates": [
```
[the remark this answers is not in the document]
idavis: @vxfield  I don't know. It is all inherited config.

idavis:
Supersedes #409

idavis:
Closes #410

PR #421, opened 2023-02-27
Microsoft QC provider and estimator target

msoeken:
This creates dedicated classes for the `microsoft-qc` provider targets (`microsoft.estimator`) with corresponding job and result objects.


azure-quantum/azure/quantum/target/target.py
```diff
@@ -31,8 +31,9 @@ def __init__(
         provider_id: str = "",
         content_type: ContentType = ContentType.json,
         encoding: str = "",
-        average_queue_time: float = None,
-        current_availability: str = ""
+        average_queue_time: Union[float, None] = None,
+        current_availability: str = "",
+        job_cls: Type[Job] = Job
```
xinyi-joffre, line 36: Can you add some comments to this class description around why these changes to Job/BaseJob are needed?

msoeken: Added documentation. Instead of adding it to `__init__` I can create a method to `Target` that the derived target can override.

azure-quantum/tests/unit/test_microsoft_qc.py
```diff
@@ -0,0 +1,70 @@
...
+                self.mock_wait(Job.wait_until_completed)
+            ):
+                ws = self.create_workspace()
+                estimator = ws.get_targets("microsoft.estimator")
+
+                ccnot = self._ccnot_bitcode()
+                job = estimator.submit(ccnot)
+                assert type(job) == MicrosoftEstimatorJob
+                job.wait_until_completed()
+                result = job.get_results()
```
xinyi-joffre, line 68: Should we add assert that job is successful before getting results? If error, maybe print job id and error

PR #428, opened 2023-03-15
Data structure for input params and entry point arguments

msoeken:
Creates a dedicated `InputParams` object that can be used as `input_params` as an alternative to a `dict`. The object is created with a `make_params` method on the target class and contains target parameter validations. The `InputParams` also has an `argument` field to assign QIR entry point arguments easily.


azure-quantum/azure/quantum/argument_types/types.py
```diff
@@ -0,0 +1,36 @@
...
+    @property
+    def value(self):
+        if self.step is None:
+            return {"start": self.start, "end": self.end}
+        else:
+            return {"start": self.start, "end": self.end, "step": self.step}
+
+@dataclass
+class EmptyArray:
+    type: Type
```
msoeken, line 36: Rename it to `element_type`

azure-quantum/tests/unit/test_microsoft_qc.py
```diff
@@ -115,5 +116,81 @@ def test_estimator_failing_job(self):
...
+    def test_estimator_params_validation(self):
+        """
+        Checks validation cases for reource estimation parameters.
+        """
+        params = MicrosoftEstimatorParams()
+
+        params.error_budget = 0.1
+        params.as_dict()
+
+        params.error_budget = 2
```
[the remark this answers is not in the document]
msoeken: Addressed in commit.

azure-quantum/azure/quantum/target/target.py
```diff
@@ -149,7 +150,10 @@ def submit(
         :return: Azure Quantum job
         :rtype: Job
         """
-        input_params = input_params or {}
+        if isinstance(input_params, InputParams):
+            input_params = input_params.as_dict()
+        else:
```
ashwinmayya, line 155: There's some changes here from XField. You might need to resolve some merge conflicts.

azure-quantum/azure/quantum/target/params.py
```diff
@@ -0,0 +1,243 @@
...
+
+        return result
+
+
+class InputParams(InputParamsItem):
+    """
+    Class to define input parameters.
+
+    This class allows to define inout parameters for non-batching and batching
+    jobs.  The instance represents a batching job, if and only if num_items is
```
ashwinmayya, line 179: I don't think we expose the notion of a "batching job" anywhere externally, so I think it is best to avoid any references in documentation

vxfield (approved):

azure-quantum/azure/quantum/target/target.py
msoeken: I paid attention during merging.

azure-quantum/azure/quantum/target/params.py
msoeken: We do expose it here in the documentation: https://learn.microsoft.com/en-us/azure/quantum/how-to-work-with-re#batching-with-q

PR #432, opened 2023-03-20
Allow the submission of `QSharpCallable` as QIR from the `qsharp` package

vxfield:
This PR enables the submission of `qsharp.QSharpCallable` as QIR via `azure-quantum`.

### Example:
```python
import qsharp
GenerateRandomBit_Inline = qsharp.compile("""
    open Microsoft.Quantum.Intrinsic;
    operation GenerateRandomBit_Inline() : Result {
        use q0 = Qubit();
        H(q0);
        return M(q0);
    }
""")

job = target.submit(input_data=GenerateRandomBit_Inline, name="Job 1")
```

> Note: although we are introducing the dependency of `qsharp` into `azure-quantum` in our QDK build pipeline for integration testing purposes, the dependency is completely optional for end-users.
The `azure-quantum` package will first check if the type name of a `target.submit()` input is `QSharpCallable` (as a string), and only if that's the case it will attempt to call the `_repr_qir_()` method from the `QSharpCallable`. So, the dependency will only realize when the user already had the `qsharp` package needed to create an `QSharpCallable` and pass it to the `target.submit()`.


azure-quantum/azure/quantum/target/target.py
```diff
@@ -132,7 +132,22 @@ def submit(
...
         """
+
         input_params = input_params or {}
+
+        # If the input_data is `QSharpCallable` (coming from the IQ# `qsharp` Python Package)
+        # we need to convert it to QIR bitcode and set the necessary parameters for a QIR job.
+        if (input_data and type(input_data).__name__ == "QSharpCallable"):
+            kwargs["input_data_format"] = kwargs.get("input_data_format", "qir.v1")
+            kwargs["content_type"] = kwargs.get("content_type", "qir.v1")
+            kwargs["output_data_format"] = kwargs.get("output_data_format", "microsoft.quantum-results.v1")
```
ashwinmayya, line 143: NIT: Adding to kwargs makes it a bit hard to read IMO. It may be better to set input_data_format, output_data_format and content_type directly in lines 141-143, and wrap lines 151-153 within an else block for non-Q#callable paths.

azure-quantum/azure/quantum/target/target.py
```diff
@@ -132,7 +132,22 @@ def submit(
         :return: Azure Quantum job
         :rtype: Job
         """
+
         input_params = input_params or {}
+
+        # If the input_data is `QSharpCallable` (coming from the IQ# `qsharp` Python Package)
+        # we need to convert it to QIR bitcode and set the necessary parameters for a QIR job.
+        if (input_data and type(input_data).__name__ == "QSharpCallable"):
```
ashwinmayya, line 140: The description says that we will attempt a import qsharp if type is QsharpCallable, but I don't see that happening here. Is that intentional?

azure-quantum/azure/quantum/target/quantinuum.py
```diff
@@ -72,14 +65,19 @@ def submit(
         :return: Azure Quantum job
```
ashwinmayya, line 65: NIT: Update comment on line 55 since it can now accept qir

azure-quantum/azure/quantum/target/quantinuum.py
```diff
@@ -72,14 +65,19 @@ def submit(
...
         :rtype: Job
         """
+        input_data = kwargs.pop("input_data", circuit)
+        if input_data is None:
+            raise ValueError(
+                "Either the `circuit` parameter or the `input_data` parameter must be have a value."
```
ashwinmayya, line 71:
```suggestion
                "Either the `circuit` parameter or the `input_data` parameter must have a value."
```

azure-quantum/azure/quantum/target/target.py
vxfield: I updated the description. The import is not needed.

azure-quantum/azure/quantum/target/quantinuum.py
vxfield: Fixed

azure-quantum/azure/quantum/target/quantinuum.py
vxfield: When talking to Cesar, we decided to minimize the exposure of this functionality everywhere.

azure-quantum/azure/quantum/target/target.py
vxfield: Fixed.

azure-quantum/tests/unit/test_qsharp.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.mark.qsharp
+@pytest.mark.live_test
+class TestQSharpQIRJob(QuantumTestBase):
+
+    @pytest.mark.rigetti
+    def test_qsharp_qir_rigetti(self):
+        self._run_job("rigetti.sim.qvm")
+
+    @pytest.mark.quantinuum
```
ashwinmayya, line 20: Should we remove this now?

vxfield: removed

PR #431, opened 2023-03-20
Session support for `azure-quantum` Python SDK

vxfield:
This PR adds support for Sessions via the `azure-quantum `Python SDK, allowing customers to open, close and list sessions, and submit jobs and list jobs as children of a session.

### Example:
```python
with backend.open_session(name="Qiskit Session") as session:
    job1 = backend.run(circuit=circuit, shots=100, job_name="Job 1")
    job2 = backend.run(circuit=circuit, shots=100, job_name="Job 2")
    job3 = backend.run(circuit=circuit, shots=100, job_name="Job 3")
```

PR #435, opened 2023-03-21
Add documentation and tests for sessions

vxfield:
Continuation of #431


azure-quantum/azure/quantum/job/session.py
```diff
@@ -3,27 +3,53 @@
...
+    :param workspace: Workspace instance to open the session on
     :type workspace: Workspace
-    :param job_details: Job details model,
-            contains Job ID, name and other details
-    :type job_details: JobDetails
+
+    :param details: Session details model, containing the session id,
+                    name, job_failure_policy, provider_id and target.
+                    Either this parameter should be passed containing all
+                    the session detail values, the same values should be
```
ashwinmayya, line 29:
```suggestion
                    the session detail values, or the same values should be
```

azure-quantum/azure/quantum/job/session.py
```diff
@@ -40,13 +66,13 @@ def __init__(
...
-            (job_failure_policy is not None)):
-            raise ValueError("""If `session_details` is passed, you should not pass `target`, 
+        if ((details is not None)
+            and ((isinstance(target, str)) or
+                 (provider_id is not None) or
+                 (id is not None) or
+                 (name is not None) or
+                 (job_failure_policy is not None))):
+            raise ValueError("""If `session_details` is passed, you should not pass `target`,
                                 `provider_id`, `session_id`, `session_name` or `job_failure_policy`.""")
```
ashwinmayya, line 76:
```suggestion
                                `provider_id`, `id`, `session_name` or `job_failure_policy`.""")
```

ashwinmayya, line 76:
```suggestion
                                `provider_id`, `session_id`, `name` or `job_failure_policy`.""")
```

azure-quantum/azure/quantum/job/session.py
```diff
@@ -73,54 +99,124 @@ def __init__(
...
+        :rtype: Session
+        """
         self.workspace.open_session(self)
         return self
 
     def close(self) -> "Session":
+        """Closes a session, not allowing further jobs to be submitted under
+           the session.
+
+        :return: The session object with updated details after its closing.
```
ashwinmayya, line 138: Does the autogenerated client take care of updating itself based on response?

azure-quantum/azure/quantum/job/session.py
```diff
@@ -139,6 +235,37 @@ def open_session(
         job_failure_policy: Union[str, SessionJobFailurePolicy, None] = None,
         **kwargs
     ) -> Session:
+        """Opens a session and associates all future job submissions to that
+           session until the session is closed (which happens automatically
+           after existing a `with` block).
```
ashwinmayya, line 240:
```suggestion
           after exiting a `with` block).
```

azure-quantum/azure/quantum/job/session.py
```diff
@@ -139,6 +235,37 @@ def open_session(
         job_failure_policy: Union[str, SessionJobFailurePolicy, None] = None,
         **kwargs
     ) -> Session:
+        """Opens a session and associates all future job submissions to that
```
ashwinmayya, line 238: Could you document what happens if a session is already started?

azure-quantum/azure/quantum/job/session.py
```diff
@@ -139,6 +235,37 @@ def open_session(
...
+
+        :param job_failure_policy: The policy that determines when a session would fail,
+                                close and not accept further jobs.
+        :type job_failure_policy: Union[str, SessionJobFailurePolicy, None]
+
+        :return: The session object with updated details after its opening.
+        :rtype: Session
+        """
         if self.latest_session:
             self.latest_session.close()
```
ashwinmayya, line 270: The service already handles gracefully the case where close is issued on a session that is already in a terminal state, but it would be  a nice optimization to avoid making the call if the client knows that it is already closed.

azure-quantum/azure/quantum/workspace.py
```diff
@@ -375,8 +374,8 @@ def list_sessions(
     ) -> List[Session]:
         """Get a list sessions for the given workspace.
```
ashwinmayya, line 375:
```suggestion
        """Gets the list of sessions in the given workspace.
```

azure-quantum/azure/quantum/workspace.py
```diff
@@ -409,23 +418,45 @@ def refresh_session(
...
+
+        :param session: The session to be refreshed.
+        :type session: Session
+        """
         session.details = self.get_session(session_id=session.id).details
 
     def get_session(
         self,
-        session_id:str
+        session_id: str
```
ashwinmayya, line 431: Ideally the session method would just take id as the param instead of session_id. But I see that the job methods take job_id so may be we keep it as is for consistency.

azure-quantum/azure/quantum/job/session.py
vxfield: the autogenerated client does not, but the `workspace.close_session()` does
```
session.details = client.close(session_id=session.id)
```

azure-quantum/azure/quantum/workspace.py
vxfield: Yes, that's why I kept it like this. Also it makes it clear that this is the session's id (since we are in the workspace object).
The session object does not have the `session_` prefix.

azure-quantum/azure/quantum/job/session.py
vxfield: Done. Handled in the `workspace.close_session`.

ashwinmayya (approved):

azure-quantum/tests/unit/test_session.py
```diff
@@ -0,0 +1,285 @@
...
+        self.pause_recording()
+
+    @pytest.mark.skip()
+    def test_session_job_qiskit(self):
+        self.pause_recording()
+
+        from azure.quantum.qiskit import AzureQuantumProvider
+        workspace = self.create_workspace()
+        provider = AzureQuantumProvider(workspace=workspace)
+        backclose = provider.get_backclose("ionq.simulator")
```
ashwinmayya, line 267: Looks like a replace all error :)
provider.backend(..)

PR #447, opened 2023-03-27
Setting header name for count extraction in qiskit results.

idavis:
Fixes QIR jobs failing to get counts due to missing metadata header in result object.

```python
job = backend.run(circuit)
result = job.result()
print(result.get_counts(0))
print(result.get_counts())
print(result.get_counts(circuit))
```

This also effects `.data()`, `.memory()`, and others.


vxfield (approved):

azure-quantum/tests/unit/test_qiskit.py
```diff
@@ -898,7 +898,11 @@ def test_qiskit_submit_to_rigetti(self):
 
             if JobStatus.DONE == qiskit_job.status():
                 result = qiskit_job.result()
-                print(result)
+                # verify we can get the counts with the circuit and without
+                # These will throw if job metadata is incorrect
+                _ = result.get_counts(circuit)
```
vxfield, line 903: Consider using `self.AssertNotNone` or `self.AreEqual` if possible.

PR #459, opened 2023-04-24
[DRAFT] Additional tests for Sessions, and improvements/fixes for the testing framework

vxfield:
This PR has 4 main updates (which will be broken down into 3 smaller PRs):

1) **New test cases for Sessions**
This was the original goal of this PR. The changes can be seen in [azure-quantum/tests/unit/test_session.py](https://github.com/microsoft/qdk-python/compare/xfield/session4?expand=1#diff-135cfd07be23dc2c65dac58a7aa7fae8912bca625730f4f568b0f86c79318eeb) file.

2) **Make randomly created  job-ids and session-ids have a deterministic value in test recordings**
We had a limitation that was not permitting multiple jobs to be submitted and recorded/played-back via the testing framework.
The improvement was mainly to automatically detect newly created random job-ids and session-ids in HTTP requests and replace them with a deterministic sequencial id that could be used during playback.
These changes can be seen in [azure-quantum/tests/unit/common.py](https://github.com/microsoft/qdk-python/compare/xfield/session4?expand=1#diff-3bc780450a034e3690c34afd809057e127772b1cb8a5a848659622a4c661d344), and were accompanied by some refactor of other test cases that relied on job-ids.

3) **Consistently pass/receive `provider_id` parameters in the Target constructors**
There was a limitation/bug in our dynamic instantiation of Targets, in which provider_ids were not being consistently passed to the constructors of concrete Target types.
I had to do some updates in all Target classes to make sure they would handle a `provider_id` parameter (that could come via `kwargs`) in a consistent manner.

4) **New test recordings for all tests**
Because of (2), all test recordings had to be re-recorded with the new deterministic ids.


azure-quantum/azure/quantum/target/ionq.py
```diff
@@ -54,6 +54,8 @@ def __init__(
         encoding: str = "",
         **kwargs
     ):
+        provider_id = kwargs.pop("provider_id", provider_id)
```
ashwinmayya, line 57: I don't quite understand why we need to make each target support overriding the provider_id in the kwargs. I imagine we are only doing this to support the echo targets?

PR #464, opened 2023-05-12
Pin urllib3 for E2E tests

achocron:
vcrpy is incompatible with urllib3 2.0.2, so this pins urllib3 to an older version.

### DIFF
--- a/azure-quantum/tests.live/Run.ps1
+++ b/azure-quantum/tests.live/Run.ps1
@@ -105,6 +105,9 @@ if ($PackageDir -Match "azure-quantum") {
     Copy-Item -Path (Join-Path $PackageDir "tests" "unit" "*.qs") -Destination $PSScriptRoot
 }
 
+# Pin urllib3 for compatibility with vcrpy
+pip install urllib3==1.26.15
+
 python -m pytest -v `
     --junitxml=junit/test-results.xml `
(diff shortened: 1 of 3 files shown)


azure-quantum/tests.live/Run.ps1
```diff
@@ -105,6 +105,10 @@ if ($PackageDir -Match "azure-quantum") {
     Copy-Item -Path (Join-Path $PackageDir "tests" "unit" "*.qs") -Destination $PSScriptRoot
 }
 
+# Pin urllib3 for compatibility with vcrpy
+Write-Host "##[info]Install version of urllib3 that is compatible with vcrpy"
+pip install urllib3==1.26.15
```
ashwinmayya, line 110: Could this be in the requirements.txt file instead?

ashwinmayya: Or some other requirements file that is used during build

shenyingjun (approved):

azure-quantum/tests.live/Run.ps1
achocron: The requirements files seem to apply to not just test environments. This seemed to by the only test-specific place to include the pin.

vxfield:
Closed as superseded by #468

PR #468, opened 2023-05-23
Pin urllib3 to fix issue with vcrpy

vxfield:
Replaces (and based on) #464.

`vcrpy` is incompatible with the latest `urllib3` `2.0.2` (https://github.com/kevin1024/vcrpy/issues/688).

This PR is a temporary workaround to pin some older versions of the dependencies to unblock our test pipelines.

We take an indirect dependency on `vcrpy` and `urllib3` from the `azure-devtool` package, which contains the Azure SDK test infrastructure.
Once `vcrpy` is fixed and `azure-devtool` picks the fix, we can undo this workaround.

### DIFF
--- a/azure-quantum/requirements-dev.txt
+++ b/azure-quantum/requirements-dev.txt
@@ -1,2 +1,9 @@
-azure-devtools>=1.2.0,<2.0
+# The following lines are to be removed 
+# after https://github.com/kevin1024/vcrpy/issues/688 is fixed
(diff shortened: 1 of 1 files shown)


kuzminrobin (approved):
Looks good to me. Once the `qdk.release` passes the step fixed by this change (see earlier failure in "Build QDK-Python"/"Test all") feel free to merge (the subsequent steps may still fail and need a separate fix).

PR #471, opened 2023-05-25
Let E2E tests use the latest released package

vxfield:
This PR checks for the environment variable `PICK_QDK_VERSION` that is used by the E2E live test pipeline and when it's set to "auto" (which is the default mode), it will install the latest published `azure-quantum` package from PyPI instead of using the build artifact.

This allows the E2E test pipeline to use the latest test code from the QDK-Python `main` branch while testing against the latest released `azure-quantum` package.

### DIFF
--- a/azure-quantum/tests.live/Install-Artifacts.ps1
+++ b/azure-quantum/tests.live/Install-Artifacts.ps1
@@ -23,7 +23,7 @@ if (-not $Env:PYTHON_OUTDIR) {
     "== We will install $PackageName from source." | Write-Host
     "" | Write-Host
 
-    Install-PackageInEnv -PackageName $PackageName -FromSource $True
+    Install-PackageInEnv -PackageName "$PackageName[all]" -FromSource $True
 
     "" | Write-Host
     "== $PackageName installed from source. ==" | Write-Host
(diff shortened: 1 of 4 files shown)


azure-quantum/tests.live/Install-Artifacts.ps1
```diff
@@ -34,13 +34,18 @@ if (-not $Env:PYTHON_OUTDIR) {
     "== To use build artifacts, download the artifacts locally and point the variable to this folder." | Write-Warning
     "" | Write-Warning
     Exit 1
+} 
+# this condition is used by the E2E Live test pipeline
+elseif (($Env:PICK_QDK_VERSION -eq "auto") -and ([string]::IsNullOrEmpty($PackageName) -or ($PackageName -eq "azure-quantum"))) {
```
ashwinmayya, line 39: Could you also add a comment explaining the second clause in the elseif?

ashwinmayya (approved):

azure-quantum/tests.live/Install-Artifacts.ps1
vxfield: That condition was not really necessary here so I removed it.

PR #479, opened 2023-06-29
Deprecate Microsoft QIO and 1QBit

frtibble:
This PR:
- removes references to Microsoft QIO targets from the Microsoft provider, including Microsoft specific functionality such as protobuf support, and squared linear combination (SLC) terms, npz functionality, and `HardwarePlatform` (which supported CPU/FPGA solvers, FPGA has been deprecated and CPU is soon to be).
- removes references to 1QBit


vxfield (commented):
I think we should also delete the test recording files for the test cases that we are removing,

vxfield (approved):

PR #515, opened 2023-10-27
Deprecate qdk.chemistry and rewrite README

xinyi-joffre:
This PR contains a refresh of the README to clarify that this repository is primarily for the azure-quantum python SDK.

It also adds deprecation notices for qdk.chemistry and consolidates the dev setup information in CONTRIBUTING.md.


CONTRIBUTING.md
```diff
@@ -3,6 +3,14 @@
...
 follow the instructions provided in [Microsoft Azure Projects Contribution Guidelines](https://opensource.microsoft.com/collaborate/).
 
+## Pre-requisites
+
+Install pre-reqs:
+
+```bash
+pip install azure_devtools pytest pytest-azurepipelines pytest-cov
+```
+
```
kikomiss, line 13: Would you think we could add `pytest-azurepipelines` and `pytest-cov` into `requirements-dev.txt` file so that these packages get installed automatically when we create the `azurequantum` Conda environment?
```bash
conda env create -f environment.yml
```
We already seem to install dev packages [this way](https://github.com/microsoft/qdk-python/blob/68a3fd28e1f62c0d098afa15dd03ad08d13d006a/azure-quantum/requirements-dev.txt#L2C1-L2C27)

xinyi-joffre: That makes sense! Can we do that changes as a separate PR in case it causes breakages in pipelines?

kikomiss: Discussed offline. Will consider adding the dependencies into `requirements-dev.txt` in a separate PR

README.md
```diff
@@ -1,70 +1,33 @@
 [![Build Status](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_apis/build/status/microsoft.qdk-python?branchName=main)](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/latest?definitionId=32&branchName=main)
 
-# QDK-Python
+# Azure Quantum SDK
 
 ## Introduction
 
-QDK-Python is the repository for Python packages of the Quantum Development Kit (QDK). Currently, this consists of the following packages:
+This repository contains the azure-quantum python SDK.
```
minestarks, line 7: nit: capitalization
```suggestion
This repository contains the azure-quantum Python SDK.
```

visualization/README.md
```diff
@@ -1,6 +1,6 @@
 # Visualization
 
-This folder includes the visualization project.
+This folder includes the visualization project. This contains visualizations for the resource estimation feature.
```
minestarks, line 3: Nit: repetitive wording
```suggestion
This folder contains visualizations for the resource estimation feature.
```

qdk/README.md
minestarks, line 1: Would we want to display a deprecation warning during install as well?

minestarks (approved):

qdk/README.md
xinyi-joffre: I tried displaying a deprecation warning during install in setup.py, but looks like pip suppresses warnings unless you fail the install entirely. It only shows up if you use the -v flag during pip install, and even then it's buried in a wall of install logs.

The other option is to add it to the __init__.py during `import qdk`, but if we set level to DeprecationWarning, it's also suppressed by default unless the user runs python with -w flag. We could emit it as UserWarning level, which shows up slightly better.

Do you have a preference or suggestions on other workarounds for deprecation warnings?

minestarks: The setup log would have been my suggestion, but I can see how that's not very helpful if it's going to be buried in logs. Still, better than nothing I guess? I'm okay with whichever way you choose to go, given that none of the solutions seem ideal.

xinyi-joffre: It's buried in logs, and also requires that user install using verbose flag "pip install azure-quantum -v" for instance

kikomiss: > I tried displaying a deprecation warning during install in setup.py, but looks like pip suppresses warnings unless you fail the install entirely. It only shows up if you use the -v flag during pip install, and even then it's buried in a wall of install logs.
> 
> The other option is to add it to the **init**.py during `import qdk`, but if we set level to DeprecationWarning, it's also suppressed by default unless the user runs python with -w flag. We could emit it as UserWarning level, which shows up slightly better.
> 
> Do you have a preference or suggestions on other workarounds for deprecation warnings?

I think we could add 
```python
warn(f"The module {__name__} is going to be deprecated...", DeprecationWarning, stacklevel=2)
```
to `azure-quantum/azure/quantum/__init__.py`. The `DeprecationWarning` gets displayed when user runs code using the deprecated module and since this is the root module, user will see the warning message everytime they use `azure.quantum` namespace. 
image
The warning will be suppressed only if user passes `-Wignore` flag.

xinyi-joffre: Thanks, I had added it to both places this morning, but let me update the type to DeprecationWarning and add stacklevel=2, which is a better type than the default UserWarning

xinyi-joffre: Fixed

PR #524, opened 2023-11-08
Remove RigettiTarget.ASPEN_M_3

tncc:
Remove RigettiTarget.ASPEN_M_3, replace in testing with ANKAA_2.  Re-run tests to generate new test data.


xinyi-joffre:
Let's hold this PR until we hear back on if Aspen M3 is really decommissioning. Sounds like it may be coming back up with 8 qubits.

masenol:
Checked with Fabrice, docs are updated with this message and target is back Available. We can abandon this PR.

04diiguyi:
Abandon this PR because Aspen M3 is back

PR #528, opened 2023-11-21
supporting status attribute and status dictionary item

ivanbasov:


msoeken (approved):

xinyi-joffre (approved):

azure-quantum/azure/quantum/target/microsoft/result.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, data: Union[Dict, List]):
             super().__init__(data)
 
             self._is_simple = True
-            if hasattr(self, 'status') and self['status'] == "success":
+            if self.__is_succeeded():
```
xinyi-joffre, line 41: I think single underscore beforehand is usually more standard for internal functions, but I'm not a python expert, so could be wrong!

xinyi-joffre: I guess it can mean a class private variable that can't be overridden by subclasses? In that case, it may make sense, and feel free to resolve this comment!

ivanbasov: Thank you, @xinyi-joffre ! Let us use a single one.

PR #529, opened 2023-11-21
support EstimationResult and dictionary items

ivanbasov:


azure-quantum/azure/quantum/target/microsoft/result.py
```diff
@@ -18,6 +18,8 @@ def __init__(self, content: str):
     def _repr_html_(self):
         return self.content
 
+def is_succeeded(obj):
```
xinyi-joffre, line 21: It feels a little random to put this helper function in the middle of the file.

Can we consider moving to top of the file above the classes?

I think the standard in python is still to annotate with an underscore, if it's meant to be private function not for export. Can we add back the initial underscore? Also, consider adding it as a static function on a class instead of standalone private function.

This module also seems to have a number of definitions not meant for export, and is_succeeded is a very general name. Can we add an `__all__ = ['MicrosoftEstimatorResult']` to the top of the file (and any other export needed other than MicrosoftEstimatorResult if exporting as wildcard), so that someone doesn't import conflicting functions/classes by default.

xinyi-joffre (approved):

PR #536, opened 2023-12-12
Add initial samples for gallery

xinyi-joffre:
Migrate select azure-quantum SDK notebook samples from the https://github.com/microsoft/quantum repository that will also be available as part of hosted notebooks.


samples/hello-world/HW-ionq-qsharp.ipynb
```diff
@@ -0,0 +1,381 @@
...
+      "id": "b9d29721-e717-480e-8f33-308682bacdd1"
+    },
+    {
+      "cell_type": "markdown",
+      "source": [
+        "### 2. Build the quantum program\n",
+        "\n",
+        "Let's create a simple Q# program to run.\n",
+        "\n",
+        "First, let's initialize the Q# environment and set the target profile to Base Profile. Today, Q# compilation for Azure Quantum targets only supports the Base Profile, a subset of all Q# commands."
```
kikomiss, line 143: Just to consider if it would make more sense to say:
"Today, Azure Quantum targets only support the Base Profile, a subset of all Q# commands."
otherwise, it might sound like we own Q# language, but cannot compile any program.

kikomiss (approved):

PR #540, opened 2023-12-20
feature: support for microsoft.dft target

kikomiss:
Usage: 
```python
workspace = Workspace(resource_id = workspace_resouce_id, location = workspace_location)

target = workspace.get_targets("microsoft.dft")

dft_input_params = {...}

with open("./molecule.xyz", "r") as file:
    job = target.submit(input_data=file.read(), input_params=dft_input_params)
    results = job.get_results()
    print(results[0]["return_result"])
```


azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
```diff
@@ -0,0 +1,56 @@
...
+        try:
+            job_results = super().get_results(timeout_secs)
+            return job_results["results"]
+        except RuntimeError:
+            if self.details.status == "Failed":
+                job_blob_properties = self.download_blob_properties(self.details.output_data_uri)
+                if job_blob_properties.size > 0:
+                    job_failure_data = self.download_data(self.details.output_data_uri)
+                    raise DftRuntimeError("An error occurred during DFT-job execution. " \
+                                          "Use error.get_details() for more information.", job_failure_data)
```
ashwinmayya, line 24: Maybe move some of this logic to a base class so it can be reused across derived classes. Something like _get_failed_output_or_raise() that returns either what job_failure_data expects or throws

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
```diff
@@ -0,0 +1,56 @@
...
+            raise
+
+
+class DftRuntimeError(RuntimeError):
```
ashwinmayya, line 28: Consider having a base class like JobError so we can establish a pattern across targets.

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
```diff
@@ -0,0 +1,56 @@
...
+                    and len(self._details["results"]) > 0 \
+                    and "error" in self._details["results"][0] \
+                    and "error_message" in self._details["results"][0]["error"]:
+                self._message = f'{message} Inner error: {self._details["results"][0]["error"]["error_message"]}' 
+        except:
+            self._details = details
+            self._message = message
+
+
+    def get_details(self) -> Any:
```
ashwinmayya, line 48: Would get_failed_result be a better name?

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
ashwinmayya: or maybe FailedJobWithResultError since we seem to prefer Result over Output

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
kikomiss: Thanks for the suggestion. Moved the logic for getting failure results to the base `Job` class. Defined a "classmethod" `_allow_failure_results()` to allow derived Job-classes to define the failure behavior for themselves.

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
kikomiss: Renamed the error-class to `JobFailedWithResultsError` as the method producing this error is called `get_results()` I thought making "results" plural would align.

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
kikomiss: Renamed to `get_failure_results()` as I thought it would sound mode straightforward, but open for changing it to `get_failed_results()`

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
```diff
@@ -0,0 +1,34 @@
...
+        super().__init__(workspace, job_details, **kwargs)
+
+
+    def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT) -> Dict[str, Any]:
+        try:
+            job_results = super().get_results(timeout_secs)
+            return job_results["results"]
+        except JobFailedWithResultsError as e:
+                failure_results = e.get_failure_results()
+                if "results" in failure_results \
```
ashwinmayya, line 21: Does this line 21 cause any issues if failure_results is a string instead of a dict?

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
```diff
@@ -0,0 +1,34 @@
...
+                    and len(failure_results["results"]) > 0 \
+                    and "error" in failure_results["results"][0] \
+                    and "error_message" in failure_results["results"][0]["error"]:
+                    message = f'{e.get_message()} Inner error: {failure_results["results"][0]["error"]["error_message"]}'
```
ashwinmayya, line 25: Could we also add error_type in the message?

ashwinmayya (commented):
Left some comments

azure-quantum/azure/quantum/job/job.py
```diff
@@ -130,3 +137,16 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
         except:
             # If errors decoding the data, return the raw payload:
             return payload
+
+
+    @classmethod
+    def _allow_failure_results(self) -> bool:
```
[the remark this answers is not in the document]
kikomiss: Good question, I think we have a pattern in Target class where we have `_get_job_class()` method returning Job type that can be overridden in subclasses. I would suggest to follow that pattern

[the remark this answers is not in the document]
ashwinmayya: Ok

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -269,23 +270,26 @@ def download_data(self, blob_uri: str) -> dict:
...
+    def download_blob_properties(self, blob_uri: str):
+        """Download Blob properties
+
+        :param blob_uri: Blob URI
+        :type blob_uri: str
+        :return: Blob properties
+        :rtype: dict
+        """
+
+        lob_uri_with_sas_token = self._get_blob_uri_with_sas_token(blob_uri)
```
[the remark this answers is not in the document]
kikomiss: good catch! fixed

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
kikomiss: Added the `error_type` to the message. 
Example: 
"An error occurred during job execution. Error type: convergence_error. Message: SCF failed to converge after 1 steps"

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
kikomiss: Not this particular line since Python treats 
```python
 str1 in str2
```
as 
```python
str2.contains(str1)
```
but it could've very well failed in the next condition
```python
failure_results["results"]
```
if `failure_results` is a `str`. Added extensive validation to prevent error. Thanks!

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
```diff
@@ -0,0 +1,46 @@
...
+        """
+        return True
+
+
+    @staticmethod
+    def _is_dft_failure_results(failure_results: Union[Dict[str, Any], str]) -> bool:
+         return isinstance(failure_results, dict) \
+                    and "results" in failure_results \
+                    and isinstance(failure_results["results"], collections.abc.Sequence) \
+                    and len(failure_results["results"]) > 0 \
```
ashwinmayya, line 41: Is there a more concise way to do this check?

azure-quantum/azure/quantum/target/microsoft/elements/dft/target.py
```diff
@@ -0,0 +1,54 @@
...
+        # make sure to not pass argument twice
+        kwargs.pop("provider_id", None)
+
+        super().__init__(
+            workspace=workspace,
+            name=name,
+            input_data_format="microsoft.xyz.v1",
+            output_data_format="microsoft.dft-results.v1",
+            provider_id="microsoft-elements",
+            content_type=ContentType.json,
```
ashwinmayya, line 38: What does this end up doing? Set content-type metadata on the blob? I think text/plain would be more appropriate for the xyz format

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT) -> Dict[str, Any]:
+        try:
+            job_results = super().get_results(timeout_secs)
+            return job_results["results"]
+        except JobFailedWithResultsError as e:
+                failure_results = e.get_failure_results()
+                if MicrosoftElementsDftJob._is_dft_failure_results(failure_results):
+                    error = failure_results["results"][0]["error"]
+                    message = f'{e.get_message()} Error type: {error["error_type"]}. Message: {error["error_message"]}'
```
ashwinmayya, line 24: Does e.get_message() add a dot (.) at the end?
Just making sure the error message looks well formatted

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
kikomiss: Unfortunately, I wasn't able to find any other type-safe way without pulling in `jsonschema` library from PyPI

azure-quantum/azure/quantum/target/microsoft/elements/dft/target.py
kikomiss: Good catch, missed this originally. Yes, this sets `Content-Type` metadata on the input data blob. Fixed.

azure-quantum/azure/quantum/target/microsoft/elements/dft/job.py
kikomiss: The original exception message ends with period, so we don't need to add it here.

ashwinmayya (approved):

xinyi-joffre (approved):

PR #542, opened 2023-12-22
Introduce the "shots" parameter to 'sumbit' function of the 'Target' object

ArthurKamalov:
This PR introduces new parameter `short` for a number of repeats for `Target.submit` function.

Things to do:

- [x] Complete changing signature for all derived classes.
- [x] Add more tests.
- [ ] Find out why the ```x-client-cpu: x64``` field is removed in the new test recordings.


azure-quantum/azure/quantum/target/microsoft/target.py
```diff
@@ -393,7 +393,12 @@ def submit(self,
                 (module, _) = to_qir_module(input_data, record_output=False)
                 input_data = module.bitcode
         finally:
-            return super().submit(input_data, name, input_params, **kwargs)
+            return super().submit(
```
kikomiss, line 396: since we're are not explicitly specify `shots` in the `submit()` function would Python make sure to pass the `shots` down to the generic `Target` as part of `**kwargs`?
Even though the generic `Target` doesn't use `shots` yet, I think it would make sense the `shots` still get passed correctly.
Could you please add unit-test for this case?

ArthurKamalov: Ah, yes, it definitely must be passed to the base submit method, in case we want to additionally handle this value in future in this base class. 
Since it is also derived from this base target, I think it just has to be explicit as anywhere else and not in `kwargs`, what do you think?

> Could you please add unit-test for this case?

I am not sure if we should add such test for a specific Target (for example this one), since I can't see much value from that, especially when it is explicitly defined.  Ideally, it would be nice to have such test, and not for a particular target but rather as a special test that exercises the "uniformity" of all existing targets. Will try to implement such test, in next commit, if possible.

azure-quantum/azure/quantum/target/ionq.py
```diff
@@ -103,13 +106,28 @@ def submit(
             )
         if input_params is None:
             input_params = {}
+
+        num_shots = kwargs.pop("num_shots", None)
+
         if num_shots is not None:
+            warn(
+                "The 'num_shots' will be deprecated. Please use 'shots' instead.",
```
kikomiss, line 114: super-nit: just to be more specific in messaging and follow the pattern defined on the line 105
```suggestion
                "The 'num_shots' parameter will be deprecated. Please, use 'shots' parameter instead.",
```

azure-quantum/azure/quantum/target/ionq.py
```diff
@@ -103,13 +106,28 @@ def submit(
...
+                category=DeprecationWarning,
+            )
+            shots = num_shots
```
kikomiss, line 117: nit: I think, I might've confused you earlier when we were discussing what happens when both `shots` and `num_shots` are specified...
Now I think that since we're deprecating the `num_shots`, the `shots` should always take precedence if `num_shots` and `shots` are specified. Could you please make this change?

azure-quantum/azure/quantum/target/microsoft/target.py
```diff
@@ -377,11 +377,32 @@ def __init__(
...
+        :param shots: Number of shots. Ignored in estimation. Defaults to None
+        :type shots: int
+        :param input_params: Input parameters
+        :type input_params: Dict[str, Any]
+        :return: Azure Quantum job
+        :rtype: Job
+        """
+
+        if shots is not None:
+            warnings.warn("The 'shots' parameter is ignored in resource estimation.")
```
kikomiss, line 404: super-nit:
```suggestion
            warnings.warn("The 'shots' parameter is ignored in resource estimation job.")
```

azure-quantum/azure/quantum/target/quantinuum.py
```diff
@@ -77,13 +80,26 @@ def submit(
...
             input_params = {}
+
+        num_shots = kwargs.pop("num_shots", None)
+
         if num_shots is not None:
+            warn(
+                "The 'num_shots' will be deprecated. Please use 'shots' instead.",
+                category=DeprecationWarning,
+            )
+            shots = num_shots
```
kikomiss, line 91: nit: as I mentioned in a comment above, could you please make a change where `shots` always take precedence over `num_shots` if both are specified? Sorry for confusion

azure-quantum/azure/quantum/target/quantinuum.py
```diff
@@ -77,13 +80,26 @@ def submit(
             )
         if input_params is None:
             input_params = {}
+
+        num_shots = kwargs.pop("num_shots", None)
+
         if num_shots is not None:
+            warn(
+                "The 'num_shots' will be deprecated. Please use 'shots' instead.",
```
kikomiss, line 88: super-nit:
```suggestion
                "The 'num_shots' parameter will be deprecated. Please, use 'shots' parameter instead.",
```

kikomiss (commented):
Looks great, thank you for working on this!

Once we add:
- Warning to `microsoft.elements.dft` target that the `shots` parameter is ignored (please, pull the latest from `main`)
- Add `shots` parameter to Qiskit backend
- Add `repetitions` parameter to `CirqTarget`
the work will be complete!

azure-quantum/azure/quantum/qiskit/backends/backend.py
```diff
@@ -46,6 +51,41 @@ def __init__(
         **fields
     ):
         super().__init__(configuration, provider, **fields)
+
```
ArthurKamalov, line 54: Since this is the most "base" backend class, we should override it with our signature with the `shots` option.

azure-quantum/azure/quantum/qiskit/backends/quantinuum.py
ArthurKamalov, line 1: @kikomiss , Note: The Quantinuum backend will not print deprecation message for passing the 'count' option because its own parameter name for shots is also `count`, so we should not prevent user from specifying it, as any other "Quntinuum-specific" option.

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
ArthurKamalov, line 1: Same as https://github.com/microsoft/azure-quantum-python/pull/542/files#r1446278629

azure-quantum/azure/quantum/target/microsoft/elements/dft/target.py
```diff
@@ -43,10 +43,17 @@ def __init__(
...
                name: str = "azure-quantum-dft-job",
+               shots: int = None,
                input_params: Union[Dict[str, Any], InputParams, None] = None,
                **kwargs) -> MicrosoftElementsDftJob:
         
-        return super().submit(input_data, name, input_params, **kwargs)
+        return super().submit(
+            input_data=input_data,
+            name=name, 
+            shots=shots,
```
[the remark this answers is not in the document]
ArthurKamalov: UPD: Added warning for ignored shots option.

azure-quantum/tests/unit/test_pasqal.py
```diff
@@ -17,47 +17,85 @@
...
+
+   def test_job_submit_pasqal_typed_input_params(self) -> None:
+       num_runs = 200
+       result = self._run_job(TEST_PULSER, InputParams(runs=num_runs))
+       self.assertIsNotNone(result)
+       self.assertEqual(sum(v for v in result.data.values() if isinstance(v, int)), num_runs)
+
+   def test_job_submit_pasqal_dict_input_params(self) -> None:
+       num_runs = 150
+       result = self._run_job(TEST_PULSER, input_params={"count": num_runs})
```
ArthurKamalov, line 55: @kikomiss, t it seems that Pascal targets now expect `count` option instead of `runs` (At least on prod).

azure-quantum/azure/quantum/target/ionq.py
ArthurKamalov: Yeah, sorry. Now that case has to be handled.

azure-quantum/azure/quantum/qiskit/backends/backend.py
```diff
@@ -114,9 +160,6 @@ def _get_input_params(self, options) -> Dict[str, Any]:
             if opt in input_params:
                 input_params[opt] = options.pop(opt)
 
-        input_params["count"] = shots_count
-        input_params["shots"] = shots_count
```
kikomiss, line 118: removing the both options seems dangerous. Did we make sure that we always specify correct property for each of the providers?

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
```diff
@@ -53,7 +55,10 @@ def __init__(
 
     @classmethod
     def _default_options(cls):
-        return Options(count=500, targetCapability="BasicExecution")
+        other_options = {
+            cls._SHOTS_PARAM_NAME: 500
```
kikomiss, line 59: we've been setting the default options through `_DEFAULT_SHOTS_COUNT` earlier, we should probably follow the same pattern here.

azure-quantum/azure/quantum/qiskit/backends/qci.py
```diff
@@ -52,7 +59,12 @@ def __init__(
 
     @classmethod
     def _default_options(cls) -> Options:
-        return Options(shots=500, targetCapability="AdaptiveExecution")
+        return Options(
+            **{
+                cls._SHOTS_PARAM_NAME: 500
```
kikomiss, line 64: we've been setting the default options through `_DEFAULT_SHOTS_COUNT` earlier, we should probably follow the same pattern here.

azure-quantum/azure/quantum/target/pasqal/target.py
```diff
@@ -103,15 +106,26 @@ def submit(
...
         :type input_params: Union[InputParams, None, Dict[str, Any]]
         :return: Azure Quantum job
         :rtype: Job
         """
+
         if isinstance(input_params, InputParams):
             typed_input_params = input_params
             input_params = {
-                "runs": typed_input_params.runs,
+                self.__class__._SHOTS_PARAM_NAME: typed_input_params.runs,
```
kikomiss, line 120: this used to be `runs` earlier and right now we seem to be setting it to `_SHOTS_PARAM_NAME = "count"` on line 67. Is this correct?

azure-quantum/azure/quantum/qiskit/backends/backend.py
ArthurKamalov: As far as I can say, everything seems ok, but I guess, we can bring that back to be on a safe side.

azure-quantum/azure/quantum/target/pasqal/target.py
ArthurKamalov: It seems so. As I noted here https://github.com/microsoft/azure-quantum-python/pull/542#discussion_r1446641584, provider does not respond on `runs` option anymore, and when job is submitted, it falls back to Pascal's default values and returns "count" in details of a result job.

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
ArthurKamalov: Yeah, I just thought it may be ok to keep it as literal since it is currently used only in one place, but I guess it wouldn't harm 🙂

azure-quantum/azure/quantum/target/pasqal/target.py
kikomiss: yep, I verified that Pasqal expects `counts`. Resolving the comment.

kikomiss (approved):

PR #554, opened 2024-01-17
Upgrading Qiskit to v0.45.*

ArthurKamalov:


azure-quantum/azure/quantum/target/quantinuum.py
```diff
@@ -160,9 +160,8 @@ def estimate_cost(
 
             else:
                 from qiskit.dagcircuit.dagnode import DAGOpNode
-                qasm = Qasm(data=circuit)
-                ast = qasm.parse()
-                dag = ast_to_dag(ast)
+                circuit_obj = loads(string=circuit)
+                dag = circuit_to_dag(circuit=circuit_obj)
```
ArthurKamalov, line 164: For now, it seems like it works even without parsing the node for the circuit program (`ast = qasm.parse()`), and just calling `circuit_to_dag` is enough for getting the same `DAGCircuit` object, but I am still not sure, so need to double-check.

azure-quantum/requirements-qiskit.txt
```diff
@@ -1,6 +1,5 @@
 qiskit-ionq>=0.3.3,<1.0
-# TODO(kuzminrobin, #494): See line below. Upon fix of #494 consider reverting fragment `<0.25.0` back to `<1.0`.
-qiskit-terra>=0.19.1,<0.25.0
 qiskit-qir>=0.4,<0.5
+qiskit>=0.45,<0.46
```
kikomiss, line 3: limiting at `0.45.*` to make sure user doesn't pull breaking changes (potentially) when 0.46 (prep for Qiskit V1) gets released. We'll make another carefully tested release when Qiskit 0.46 comes out in Feb, 2024.

xinyi-joffre (approved):

azure-quantum/requirements-qiskit.txt
xinyi-joffre, line 3: Do we need to keep this as qiskit-terra for now until qiskit-ionq and qiskit-qir makes their swap? Otherwise, we could be pulling in conflicting package from those other dependencies.

xinyi-joffre, line 3:
```suggestion
qiskit-terra>=0.45,<0.46
```

kikomiss: Reverted to use `qiskit-terra`. Thanks!

xinyi-joffre (approved):

xinyi-joffre (approved):

PR #556, opened 2024-01-23
Update quantum workspace client and add connection string support

04diiguyi:


azure-quantum/azure/quantum/_client/_client.py
```diff
@@ -53,31 +57,47 @@ class QuantumClient:  # pylint: disable=client-accepts-api-version-keyword
     :type workspace_name: str
     :param credential: Credential needed for the client to connect to Azure. Required.
     :type credential: ~azure.core.credentials.TokenCredential
-    :keyword endpoint: Service URL. Default value is "https://quantum.azure.com".
```
kashifkhan, line 56: This will be a breaking change for someone using using this kwarg. Its typical to have a deprecation warning before actually removing for folks to move over to the new api

azure-quantum/azure/quantum/workspace.py
```diff
@@ -218,6 +248,40 @@ def user_agent(self):
...
+            workspace_name = match.group(3)
+            workspace_key = match.group(4)
+            location = match.group(5)
+
+            return subscription_id, resource_group, workspace_name, location, workspace_key
+
+        raise Exception("Invalid connection string")
+
+    @classmethod
+    def from_connection_string(cls, connection_string, **kwargs):
```
kashifkhan, line 276: please type out the kwargs that are accepted here, similar to the ones in the constructor

kashifkhan: also docstrings should be here

azure-quantum/azure/quantum/workspace.py
```diff
@@ -178,6 +198,7 @@ def __init__(
         self.subscription_id = subscription_id
         self.storage = storage
         self._user_agent = user_agent
+        self.kwargs = kwargs
```
johanste, line 201: This is an old attribute. What is it supposed to be used for?

azure-quantum/azure/quantum/workspace.py
```diff
@@ -132,6 +135,7 @@ def __init__(
         location: Optional[str] = None,
         credential: Optional[object] = None,
```
johanste, line 136: This is typed way too loosely. Are you really accepting *any* kind of object here?

PR #566, opened 2024-02-01
Connection String and api-key authentication support

vxfield:
## Main changes:

1. Added support for connection string parsing and creation of a Workspace client from a connection string
2. Added support for api-key authentication
3. Added unit tests connection string and api-key authentication


azure-quantum/azure/quantum/_workspace_connection_params.py
```diff
@@ -53,13 +66,19 @@ def __init__(
...
         on_new_client_request: Optional[Callable] = None,
     ):
         self._location = None
         self._environment = None
         self._base_url = None
         self._arm_base_url = None
 
+        # connection_string is set first as it
+        # should be overridden by other parameters
+        self.apply_connection_string(connection_string)
```
04diiguyi, line 80: Just out of curiosity, what happens if we receive a valid connection string, and in the environment variables or somewhere else we set partial of the parameters. Will the final connection_parameters be a mix?

04diiguyi (dismissed):

azure-quantum/azure/quantum/_workspace_connection_params.py
vxfield: Just in the case of environment variables. Individual parameters from the connection string will be used if that parameter is not already set via other environment variables.

If the user creates the workspace from the connection string, the environment variables are not used at all.

kikomiss (approved):

04diiguyi (approved):

kikomiss (approved):

PR #574, opened 2024-02-09
fix: @pytest.mark attributes + prep for 1.1 release

kikomiss:
- removed deprecated `@pytest.mark` attributes and added missing ones
- bump to `1.1.*` version in `set_version.py` in prep for `1.1.0` release

### DIFF
--- a/set_version.py
+++ b/set_version.py
@@ -5,7 +5,7 @@
 
 import os
 
-major_minor = "1.0"
+major_minor = "1.1"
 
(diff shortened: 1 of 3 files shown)


set_version.py
```diff
@@ -5,7 +5,7 @@
 
 import os
 
-major_minor = "1.0"
+major_minor = "1.1"
```
kikomiss, line 8: I will look into baking the version into pipeline variables so that we don't need a code change when we release.

xinyi-joffre (approved):

PR #582, opened 2024-02-28
Remove versions upper-bounds limits for dependency packages

vxfield:
## Changes

This PR follows the move from the Azure SDK Core Libraries for Python to not have upper-bound version limits for the dependency packages.

## Rationale

Having upper-bound version limits for the dependency packages prevent potential and unexpected breaking changes.
However, it also prevents our customers from accessing the latest versions of these dependencies, potentially blocking them from applying security patches or integrating our libraries with other tools.

Since the `azure-quantum` package will eventually need to support newer versions as they become available (to apply security patches or unblocking customers, it's better to use the latest dependencies' versions by default and set an upper-bound version limit as a (temporary) exception / workaround.

## Tests

Our unit tests cover 77% of our code in general, but the most used scenarios and features have a code coverage of 90+%.
The CI, Release and E2E test pipelines will always install the latest dependencies and we can quickly identify breaking changes.

## In case of breaking changes

There are two potential types of breaking changes:

1) The latest version of a dependency package has a breaking change in their surface API or behavior that breaks part of the `azure-quantum` functionality.
2) The latest versions of two different packages depend on a common dependency and each of these packages impose version limits that are incompatible (for example, `PackageA` depends on `PackageC>=1.0,<1.2` and `PackageB` depends on `PackageC>=1.3` ). This will make the `azure-quantum` package un-installable because it's impossible to resolve the dependency graph.

### Mitigation 1

1) Find which dependency packages are creating the issue.
2) Identify the latest previous version that does not contain the issue.
3) Create an issue in the [azure-quantum-python](https://github.com/microsoft/azure-quantum-python/) with the workaround (ask customer to install the specific versions of dependency packages that does not cause the issue before installing the `azure-quantum` package)
4) Create issues in the repositories of the offending dependency packages (and link them to our issue)

### Mitigation 2

1) After Mitigation 1
2) Set upper-limits in the appropriate `requirements.txt` files
3) Release a new patch for the `azure-quantum` package
4) Update our issue created in Mitigation 1 with the workaround to install the latest `azure-quantum` package (which will basically do Mitigation 1 during the `azure-quantum` package installation).


kikomiss:
@vxfield , @xinyi-joffre, @ArthurKamalov 
As discussed offline we're going to keep upper bounds for dependency packages for now to avoid introducing breaking changes for customers. We could re-consider doing this at a later point.
Closing this PR as agreed.

PR #587, opened 2024-03-08
Functions to load QCSchema format into pyscf format.

ajjenkins1:
Simple functions to load QCSchema output format into one that PySCF can use. Uses PySCF functions.
Requirements: PySCF >= 2.2.1.


azure-quantum/azure/quantum/target/microsoft/elements/dft/__init__.py
```diff
@@ -2,5 +2,6 @@
 
 from .target import MicrosoftElementsDft
 from .job import MicrosoftElementsDftJob
+from .libqcschema import *
```
ashwinmayya, line 5: Kirill was mentioning that this might break customers who do not want to do the conversion and do not have pyscf installed. It might be better to put libqcshema in a separate namespace.

azure-quantum/azure/quantum/target/microsoft/elements/dft/libqcschema.py
```diff
@@ -0,0 +1,187 @@
+import json
```
kikomiss, line 1: nit: could you please also add tests to `test_microsoft_elements_dft.py` to make sure these methods return expected format. Please, let me know if you won't be able to add tests and I could take this task on.

PR #593, opened 2024-03-27
Fix expired sas token on getting job result

ArthurKamalov:
This PR add verification of the sas token's expiration time, to refresh it, if necessary.

### DIFF
--- a/azure-quantum/azure/quantum/job/base_job.py
+++ b/azure-quantum/azure/quantum/job/base_job.py
@@ -7,7 +7,8 @@ import logging
 import uuid
 
 from enum import Enum
-from urllib.parse import urlparse
+from datetime import datetime, timezone
+from urllib.parse import urlparse, parse_qs
 from typing import Any, Dict, Optional, TYPE_CHECKING
 from azure.storage.blob import BlobClient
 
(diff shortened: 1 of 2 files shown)


vxfield:
@ArthurKamalov, can you try to merge the `xfield/enable-disable-scrubbing` into your branch and use the new ability to disable the scrubbing of `se=` (example in #595).
@kikomiss: FYI

azure-quantum/tests/unit/test_job_results.py
```diff
@@ -20,6 +24,30 @@ class TestJobResults(QuantumTestBase):
...
+
+        expired_date = datetime.now(tz=timezone.utc) - timedelta(hours=240)
+        # manually create Z suffixed UTC timestamp, since Python < 3.11 can not handle it. 
+        expired_date_str = expired_date.isoformat().replace("+00:00", "Z")
+
+        query_params["se"] = expired_date_str
+        new_query_param_str = urlencode(query=query_params)
+        new_uri = f"{uri_parts.hostname}{uri_parts.path}?{new_query_param_str}"
+        job.details.output_data_uri = new_uri
+        job.get_results()
```
vxfield, line 50: I know the PR is still on draft, so I'm assuming you are going to add some asserts here later.

kikomiss (approved):

azure-quantum/tests/unit/test_job_results.py
kikomiss: Added a quick check for results existance.

vxfield (approved):

vxfield:
@kikomiss, I've made the following changes that were causing non-deterministic behavior in the test case:

1. Submit a job to the `echo-output` target instead of using the latest submitted job in the workspace.
    This guarantees deterministic behavior in terms of the job's `output_data_uri` and we can actually test the blob being downloaded (an unknown job may no longer have its output data).

2. Using the same Regular Expression that we use to sanitize test recordings to replace the `se=` query string parameter with a value formatted in the same way.

kikomiss (approved):

PR #601, opened 2024-04-16
Add new get_backends_names method for qiskit provider.

ArthurKamalov:
This method returns list of unique names of available backends of the qiskit provider.


kikomiss:
Closing PR in favor of #602

PR #602, opened 2024-04-17
Return only default backends by the Qiskit provider

ArthurKamalov:
The qiskit's 'provider.backends' method returns only default backends. 

Things to consider:
* In some cases, due to other filtering conditions, there may be resulting lists without any default backend.
  Like in this test -  https://github.com/microsoft/azure-quantum-python/blob/2fd08fe85cd90622c410707cdbe9dd8e7029ba9b/azure-quantum/tests/unit/test_qiskit.py#L573
 In such case we just do not filter for default backends and return previous version of the list.

* There wasn't any default backend for Rigetti, so I made `RigettiSimulatorBackend` default.


azure-quantum/azure/quantum/qiskit/provider.py
```diff
@@ -132,11 +116,25 @@ def backends(self, name=None, **kwargs):
...
+        default_backends = list(
+            filter(
+                lambda backend: self._match_all(
+                    backend.configuration().to_dict(), {"is_default": True}
+                ),
+                filtered_backends,
+            )
+        ) 
+        # In some cases, for example due to certain filters, there may be no default backends, 
+        # in this case return previous version of the list.
```
kikomiss, line 133: nit: The comment contradicts with the code below it. May be say something like:
If default backends were found - return them, otherwise return the `filtered_backends` collection. The latter case could happen where there's no default backend defined for the specified target.

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
```diff
@@ -94,6 +94,7 @@ def __init__(self, name: str, provider: "AzureQuantumProvider", **kwargs):
                 "open_pulse": False,
                 "gates": [{"name": "TODO", "parameters": [], "qasm_def": "TODO"}],
                 "azure": self._azure_config(),
+                "is_default": True,
```
kikomiss, line 97: I think we'll also need to add `is_default` flag onto the `RigettiQPUBackend`. As far as I see, we use `is_default` flag to deduplicate backends having the same Target name. Since we're now going to return only the default backends from `provider.backends()` we should make sure that:
- If backend has unique target name - we set `is_default=True` on it.
- If there are more than one backend with the same target name - we set `is_default=True` on only one of them.
Could you please make sure that one of the above is done for each Qiskit backend?

kikomiss (approved):

vxfield (approved):

PR #603, opened 2024-04-17
Add back the SharedTokenCacheCredential to handle token which is cached by the InteractiveBrowserCredential

ArthurKamalov:
It seems that the `InteractiveBrowserCredential` is able to save its token to persistent cache storage, but it is not able to use that cache later. ( It actually has its own method [_acquire_token_silent](https://github.com/Azure/azure-sdk-for-python/blob/6873934fb45d4b9a4ddeebdd1dcbc1e17e784a37/sdk/identity/azure-identity/azure/identity/_internal/interactive.py#L220-L235), but, as far as I can see, it does not check for token in cache, but already expects an instance of the [AuthenticationRecord ](https://github.com/Azure/azure-sdk-for-python/blob/0b9b958cd6d528d9f56a0a2c499170eb01788342/sdk/identity/azure-identity/azure/identity/_auth_record.py#L11) class, which, I assume, we should handle and store manually)
 That's why I also made an assumtion that we should use `SharedTokenCacheCredential` for that.  I know that the `SharedTokenCacheCredential` was previously removed by @vxfield and there might be a valid reason for that, but I haven't found any better workaround, so I would gland to hear whether we can use it or not.

I also found that in order to make it work on Unix, the  `azure.identity` library relies on `libsecret` which may not be installed. 
The quick workaround would be to use [unencrypted caching ](https://github.com/microsoft/azure-quantum-python/blob/72cbfdf89d36da8e41f97a14979a3bd9f5674499/azure-quantum/azure/quantum/_authentication/_default.py#L110), but I am pretty sure that's not a good idea.

### DIFF
--- a/azure-quantum/azure/quantum/_authentication/_default.py
+++ b/azure-quantum/azure/quantum/_authentication/_default.py
@@ -16,6 +16,8 @@ from azure.identity import (
     InteractiveBrowserCredential,
     DeviceCodeCredential,
     _internal as AzureIdentityInternals,
+     TokenCachePersistenceOptions,
+     SharedTokenCacheCredential, 
 )
 from ._chained import _ChainedTokenCredential
 from ._token import _TokenFileCredential
@@ -97,7 +99,18 @@ class _DefaultAzureCredential(_ChainedTokenCredential):
             credentials.append(VisualStudioCodeCredential(authority=self.authority, tenant_id=self.tenant_id))
             credentials.append(AzureCliCredential(tenant_id=self.tenant_id))
             credentials.append(AzurePowerShellCredential(tenant_id=self.tenant_id))
-            credentials.append(InteractiveBrowserCredential(authority=self.authority, tenant_id=self.tenant_id))
+            # Before trying other credential types, try to use already cached token.
+            credentials.append(SharedTokenCacheCredential(authority=self.authority))
+            credentials.append(
+                InteractiveBrowserCredential(
+                    authority=self.authority, 
+                    tenant_id=self.tenant_id,
+                    cache_persistence_options=TokenCachePersistenceOptions(
+                        # Not all linux systems has preinstalled libsecret, which is required for storage encryption. 
+                        allow_unencrypted_storage=True
+                    )
+                )
+            )
             if self.client_id:
                 credentials.append(DeviceCodeCredential(authority=self.authority, client_id=self.client_id, tenant_id=self.tenant_id))
         self.credentials = credentials


azure-quantum/azure/quantum/_authentication/_default.py
```diff
@@ -97,7 +99,18 @@ def _initialize_credentials(self):
             credentials.append(VisualStudioCodeCredential(authority=self.authority, tenant_id=self.tenant_id))
             credentials.append(AzureCliCredential(tenant_id=self.tenant_id))
             credentials.append(AzurePowerShellCredential(tenant_id=self.tenant_id))
-            credentials.append(InteractiveBrowserCredential(authority=self.authority, tenant_id=self.tenant_id))
+            # Before trying other credential types, try to use already cached token.
+            credentials.append(SharedTokenCacheCredential(authority=self.authority))
```
ArthurKamalov, line 103: On Linux, when passing `self.tenant_id` to the `SharedTokenCacheCredential`, it fails due to a mismatch of this tenant_id with tenant_id from the cache.  Still haven't figured it out.

azure-quantum/azure/quantum/_authentication/_default.py
```diff
@@ -97,7 +99,18 @@ def _initialize_credentials(self):
...
+            credentials.append(
+                InteractiveBrowserCredential(
+                    authority=self.authority, 
+                    tenant_id=self.tenant_id,
+                    cache_persistence_options=TokenCachePersistenceOptions(
+                        # Not all linux systems has preinstalled libsecret, which is required for storage encryption. 
+                        allow_unencrypted_storage=True
```
ArthurKamalov, line 110: In order to make it work on Linuux, the  `azure.identity` library relies on `libsecret` which may not be installed. 
The quick workaround would be to use unencrypted caching, but I am pretty sure that's not a good idea. 
As better option, we can pre-check for the `libsecret` on linux platform by ourselves, so it would be narrowed down to that platform.

azure-quantum/azure/quantum/_authentication/_default.py
vxfield: even if the `tenant_id` is not passed, we automatically fetch it anyway by calling [`self._discover_tenant_id_()`](https://github.com/microsoft/azure-quantum-python/blob/01c2dd05a34e5d6368e6690a279d0ccbf5a51026/azure-quantum/azure/quantum/_authentication/_default.py#L88)

ArthurKamalov: But that's for the constructor of the `_DefaultAzureCredential`, right? I meant passing the `tenant_id` to the `SharedTokenCacheCredential` itself.

vxfield: Also, since the `InteractiveBrowserCredential` already has support for caching, I'm confused why we need to add the `SharedTokenCacheCredential` in the list of credentials.

vxfield: Ok, I think I get it.
If we enable the caching in the `InteractiveBrowserCredential` it will fail even if the browser authentication would succeed but accessing the cache fails.

azure-quantum/azure/quantum/_authentication/_default.py
kikomiss: What happens if `libsecret` was not found? Would it throw an exception? If it would, it might be fine, because the `_ChainedTokenCredential` will catch it and will try the `InteractiveBrowserCredential` next.

azure-quantum/azure/quantum/_authentication/_default.py
vxfield: Please see my latest commit.
Using our own cache will mitigate issues with tokens cached by other apps.
As Kirill mentioned, if there was a change of `authority` or `tenant_id`, the `SharedTokenCacheCredential` will be skipped and the user can authenticate with the `InteractiveBrowserCredential` (which will store a new token for these new parameters).

azure-quantum/azure/quantum/_authentication/_default.py
vxfield: If `libsecret` is not found `SharedTokenCacheCredential` throws `CredentialUnavailableError` which will be caught by `_ChainedTokenCredential` as you mentioned.

vxfield:
> Observations:
It seems that the InteractiveBrowserCredential is able to save its token to persistent cache storage, but it is not able to use that cache later. ( It actually has its own method [_acquire_token_silent](https://github.com/Azure/azure-sdk-for-python/blob/6873934fb45d4b9a4ddeebdd1dcbc1e17e784a37/sdk/identity/azure-identity/azure/identity/_internal/interactive.py#L220-L235), but, as far as I can see, it does not check for token in cache, but already expects an instance of the [AuthenticationRecord ](https://github.com/Azure/azure-sdk-for-python/blob/0b9b958cd6d528d9f56a0a2c499170eb01788342/sdk/identity/azure-identity/azure/identity/_auth_record.py#L11)class, which, I assume, we have to handle and store manually)
That's why I also made an assumtion that we should use SharedTokenCacheCredential for that. I know that the SharedTokenCacheCredential was previously removed by @vxfield and there might be a valid reason for that, but I haven't found any better workaround, so I would glad to hear whether we can use it or not.

The `_auth_record` is automatically set upon successful Browser Authentication [source](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/identity/azure-identity/azure/identity/_internal/interactive.py#L176).
We don't need to worry about it if we also use the `SharedTokenCacheCredential` as you said to use the cache on a different process or instantiation of our SDK.

azure-quantum/azure/quantum/_authentication/_default.py
```diff
@@ -84,22 +87,70 @@ def _authority_or_default(self, authority: str, arm_endpoint: str):
...
+                tenant_id=self.tenant_id))
             credentials.append(AzureCliCredential(tenant_id=self.tenant_id))
             credentials.append(AzurePowerShellCredential(tenant_id=self.tenant_id))
-            credentials.append(InteractiveBrowserCredential(authority=self.authority, tenant_id=self.tenant_id))
+            # The SharedTokenCacheCredential is used when the token cache
+            # is available to attempt loading a token stored in the cache
+            # by the InteractiveBrowserCredential.
+            if cache_options:
+                credentials.append(SharedTokenCacheCredential(
+                    authority=self.authority,
```
kikomiss, line 142: I think we might also need to pass `tenant_id=self.tenant_id` to here

vxfield: done

ArthurKamalov: As I mentioned in this comment https://github.com/microsoft/azure-quantum-python/pull/603#discussion_r1569451113, on Linux I have a mismatch of a current tenant id and id from the cache, which leads to exception in the `SharedTokenCacheCredential`. Still trying to find the cause, still not sure if that only on my machine or not.

vxfield:
@ArthurKamalov, @kikomiss: I'll hand this back to you now.

I think the next steps would be to add test cases for the `_DefaultAzureCredential.get_token()` method with different authentication scenarios.

See how we instantiate the `_DefaultAzureCredential` [here](https://github.com/microsoft/azure-quantum-python/blob/01c2dd05a34e5d6368e6690a279d0ccbf5a51026/azure-quantum/azure/quantum/_workspace_connection_params.py#L409).

These new test cases:
- Can be added in the `test_authentication.py` file.
- Can mock the `get_token()` method of each credential that we support and check if `_DefaultAzureCredential.get_token()` is returning that token.
- Can mock the `_get_persistence` function to raise an Exception and check if a Warning is emitted during the `_get_cache_options()` method, and that it still returns the mocked `InteractiveBrowserCredential` token.
- etc...

azure-quantum/azure/quantum/_authentication/_default.py
```diff
@@ -84,22 +87,71 @@ def _authority_or_default(self, authority: str, arm_endpoint: str):
...
+            name="AzureQuantumSDK"
+        )
+        try:
+            # pylint: disable=protected-access
+            cache = AzureIdentityPersistentCache._load_persistent_cache(cache_options)
+            try:
+                _LOGGER.error(
+                    'Using Azure.Identity Token Cache at %s. ',
+                    cache._persistence.get_location()
+                )
```
ArthurKamalov, line 107: Sorry if I am missing something here, but should this be an error log level?

and what do we really guard here in that try/catch clause? `cache._persistence.get_location()` ?

vxfield: Good catch. I was using error log level for local testing. I updated it to info.

I guarded the `cache._persistence.get_location()` because:
1. We are calling it just for tracing / info logging purposes
2. `_persistence` is a private attribute that can change in the future
3. `get_location()` could raise an expected exception in a particular situation that we are not aware of
4. The cache implementation is different for each OS and would be hard for us to guarantee it always work
5. Finally, I thought it's not worth the risk of this info logging to break our auth :-)

But feel free to modify the code as you see better.
Like we could try to get the location, but even if fails, it we could log the rest of the info ("Using Azure.Identity Token Cache.").

vxfield: I made a small improvement based on my own comment, but feel free to adjust it.

azure-quantum/azure/quantum/_authentication/_default.py
```diff
@@ -84,22 +87,86 @@ def _authority_or_default(self, authority: str, arm_endpoint: str):
...
+                # tracing purpose.
+                _LOGGER.info(
+                    "Using Azure.Identity Token Cache at %s.",
+                    cache._persistence.get_location()
+                )
+            except: # pylint: disable=bare-except
+                _LOGGER.info("Using Azure.Identity Token Cache.")
+            return cache_options
+        except Exception as ex: # pylint: disable=broad-except
+            if (isinstance(ex, ValueError)
```
vxfield, line 114: @kikomiss, hopefully this can provide additional guidance for users facing issues with `libsecret`, but it's a best effort try, and I like your idea of having some documentation in Microsoft Learn docs.

vxfield: If we want to check if `libsecret` is working (instead of checking a `ValueError` message that can change in the future) we could call [msal_extensions.trial_run()](https://github.com/AzureAD/microsoft-authentication-extensions-for-python/blob/ae07f283bb6e7d601f5605a87be4d32b36828593/msal_extensions/libsecret.py#L110) if `sys.platform.startswith("linux")`.

vxfield: Again, I implemented my own comment :-)

azure-quantum/azure/quantum/_authentication/_default.py
ArthurKamalov: Yeah, that makes a lot of sense now, thanks :)

PR #605, opened 2024-04-29
Update cirq libraries.

ArthurKamalov:
This PR upgrades dependency libraries cirq libraries.

The codebase upgrade is almost seamless, except the handling of the newly added `get_results`  method of the `Ionq` client ([comment](https://github.com/microsoft/azure-quantum-python/pull/605/files#r1585093083))


azure-quantum/azure/quantum/cirq/targets/ionq.py
```diff
@@ -71,6 +71,14 @@ def cancel_job(self, job_id: str):
...
         azure_job = self._workspace.get_job(job_id)
         self._workspace.cancel_job(azure_job)
+    
+    def get_results(
+        self, job_id: str, sharpen: Optional[bool] = None, extra_query_params: Optional[dict] = None
+    ):
+        azure_job = self._workspace.get_job(job_id)
+        job_result = azure_job.get_results()
+        return job_result["histogram"]
+
```
ArthurKamalov, line 81: It seems that it is expected that the [get_results](https://github.com/quantumlib/Cirq/blob/7df83d7495b22f2a6eb6021af9375878451c11a2/cirq-ionq/cirq_ionq/job.py#L221) method returns histogram, so in our wrapper class we also define this method and return that histogram.

kikomiss (approved):

PR #606, opened 2024-04-30
Make get_targets return list if names is None

ArthurKamalov:
Not the `get_targets` method returns target instance instead of list only when name is not None, in other cases, it always list.

### DIFF
--- a/azure-quantum/tests/README.md
+++ b/azure-quantum/tests/README.md
@@ -62,7 +62,7 @@ See [Sequence ids](#Sequence-ids) and [Non-deterministic Job ids and Session ids
 
 To prevent potentially sensitive information to be checked-in in the repository (like authentication tokens, access keys, etc) inside the recordings, we do several text replacements in the HTTP requests and responses in the VCR pipeline before they end-up persisted in the recorded files.
 
-The [QuantumTestBase.\_\_init\_\_ method](https://github.com/microsoft/azure-quantum-python/blob/0699fa59d73a73a6fe049922349c4d626e7d4a37/azure-quantum/tests/unit/common.py#L73) contains several rules (mostly regular expressions) that are applied in the HTTP requests and reponses via several recording and playback processors that are injected in the VCR HTTP pipeline.
+The [QuantumTestBase.\_\_init\_\_ method](https://github.com/microsoft/azure-quantum-python/blob/main/azure-quantum/tests/unit/common.py#L73) contains several rules (mostly regular expressions) that are applied in the HTTP requests and reponses via several recording and playback processors that are injected in the VCR HTTP pipeline.
 We use some common processors provided by the Azure SDK framework (including AccessTokenReplacer, InteractiveAccessTokenReplacer, RequestUrlNormalizer) but we also apply custom text replacement logic in URLs and HTTP Headers via the `process_request` and `process_response` methods and some other processors/filters found at the end of the file.
 
 ### Ability to Pause Recordings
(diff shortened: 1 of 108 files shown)


azure-quantum/tests/unit/recordings/test_estimator_batching_job.yaml
```diff
@@ -26,16 +26,21 @@ interactions:
     uri: https://login.microsoftonline.com/00000000-0000-0000-0000-000000000000/oauth2/v2.0/token
   response:
     body:
-      string: '{"token_type": "Bearer", "expires_in": 1745070742, "ext_expires_in":
-        1745070742, "refresh_in": 31536000, "access_token": "PLACEHOLDER"}'
+      string: '{"error": "invalid_client", "error_description": "AADSTS7000215: Invalid
```
kikomiss, line 29: Could you please re-record this one. Seems like this failed on auth

azure-quantum/tests/README.md
```diff
@@ -76,16 +84,18 @@ if self.in_recording:
 ```
 
 ### Sequence ids
+
 By default, VCR does not allow the same request (identified by URI and HTTP Headers) to have multiple responses associated with it.
 For example, when a job is submitted and we want to fetch the job status, the HTTP request to get the job status is the same, but the response can be different, initially returning Status="In-Progress" and later returning Status="Completed".
 
-This limitation is solved by the test base class automatically injecting a `test-sequence-id` in the query string via the [CustomRecordingProcessor _append_sequence_id method](https://github.com/microsoft/qdk-python/blob/main/azure-quantum/tests/unit/common.py).
+This limitation is solved by the test base class automatically injecting a `test-sequence-id` in the query string via the [CustomRecordingProcessor \_append_sequence_id method](https://github.com/microsoft/qdk-python/blob/main/azure-quantum/tests/unit/common.py).
```
ArthurKamalov, line 91: Shouldn't it be `CustomRecordingProcessor._append_sequence_id` with more precise line URL - `https://github.com/microsoft/azure-quantum-python/blob/0699fa59d73a73a6fe049922349c4d626e7d4a37/azure-quantum/tests/unit/common.py#L458`  ?

azure-quantum/tests/README.md
```diff
@@ -76,16 +84,18 @@ if self.in_recording:
...
 This non-deterministic behavior can cause problems when the test recordings are played-back and the ids in the recordings won't match the random ids created the by SDK when the tests run again.
 
-To mitigate this problem, the `CustomRecordingProcessor` automatically looks for guids (registered with the `register_guid_regex` method) in the HTTP requests and replace them with a deterministic and sequential guid via the [CustomRecordingProcessor _search_for_guids method](https://github.com/microsoft/qdk-python/blob/main/azure-quantum/tests/unit/common.py) and the same mechanism to sanitize the recordings (regex replacement of URI, Headers and Body).
+To mitigate this problem, the `CustomRecordingProcessor` automatically looks for guids (registered with the `register_guid_regex` method) in the HTTP requests and replace them with a deterministic and sequential guid via the [CustomRecordingProcessor \_search_for_guids method](https://github.com/microsoft/qdk-python/blob/main/azure-quantum/tests/unit/common.py) and the same mechanism to sanitize the recordings (regex replacement of URI, Headers and Body).
```
ArthurKamalov, line 98: Same as https://github.com/microsoft/azure-quantum-python/pull/606/files#r1585483442

kikomiss (approved):

PR #608, opened 2024-05-10
Fix slow tests from recordings

ArthurKamalov:
This PR tries to reduce run time of tests from recordings.

1. Previously, we didn't include the request to `/.well-known/openid-configuration` during authentication, which led to additional network request for each test case that requires authentication.
2. Added more patches that put 0 seconds wait time for various polling requests.
3. Also removed all client auth secret related code in favor of Cert file.

Tests that use `Microsoft.test` provider are also skipped for until the it is fixed.


azure-quantum/tests/unit/common.py
```diff
@@ -261,9 +262,7 @@ def resume_recording(self):
 
     @property
     def is_playback(self):
-        return (self.connection_params.subscription_id == SUBSCRIPTION_ID
-                and not self.in_recording 
-                and not self.is_live)
+        return not self.in_recording and not self.is_live
```
ArthurKamalov, line 265: I've tried to check everything, and it seems that it should be safe to remove the `subscription_id` from the condition, so we consider it a playback just when the recording file is present.

azure-quantum/tests/unit/fixtures/dummy_auth_cert.pfx
kikomiss, line 1: I think we won't need this one #609 is merged. It fixes the tests in `test_authentication.py` that relied on Client Secret to be present. Could you please merge the changes from `main` once #609 is merged

PR #609, opened 2024-05-18
Remove secrets test

ashwinmayya:
Remove secrets test


azure-quantum/tests/unit/test_authentication.py
```diff
@@ -20,6 +20,7 @@
 from azure.identity import (
     CredentialUnavailableError,
     ClientSecretCredential,
+    DefaultAzureCredential,
```
kikomiss, line 23: Could you please use `_DefaultAzureCredential` that's imported below. This is our implementation based on the original `DefaultAzureCredential`

azure-quantum/tests/unit/test_authentication.py
```diff
@@ -165,8 +165,6 @@ def test_workspace_auth_default_credential(self):
             connection_params = self.connection_params
             os.environ[EnvironmentVariables.AZURE_CLIENT_ID] = \
                 connection_params.client_id
-            os.environ[EnvironmentVariables.AZURE_CLIENT_SECRET] = \
-                self._client_secret
             os.environ[EnvironmentVariables.AZURE_TENANT_ID] = \
                 connection_params.tenant_id
```
kikomiss, line 169: As far as I checked the `EnvironmentCredential` implementation if `AZURE_CLIENT_SECRET` env variables is omitted, then the `EnvironmentCredential` does not get initialized. We can remove the initialization of `AZURE_CLIENT_ID` on line 166 and `AZURE_TENANT_ID` on line 170. The `_DefaultAzureCredential` in this case should pick up AzCLI credentials and onwards.

kikomiss (approved):

masenol (approved):

PR #617, opened 2024-06-26
upgrade job.py to results v2 format by default and expose histogram/shots methods

LawsonGraham:


LawsonGraham:
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```

@microsoft-github-policy-service agree company="Microsoft"

azure-quantum/tests/unit/test_job_results.py
```diff
@@ -46,8 +46,8 @@ def test_job_get_results_with_expired_sas_token(self):
     def test_job_for_microsoft_quantum_results_v1_success(self):
         job_results = self._get_job_results("microsoft.quantum-results.v1","{\"Histogram\": [\"[0]\", 0.50, \"[1]\", 0.50]}")
         self.assertTrue(len(job_results.keys()) == 2)
-        self.assertTrue(job_results["[0]"], 0.50)
-        self.assertTrue(job_results["[1]"], 0.50)
+        self.assertEqual(job_results["[0]"], 0.50)
+        self.assertEqual(job_results["[1]"], 0.50)
```
kikomiss, line 50: good catch! could you please also do the same for all such bugs in this file like on line 56?

azure-quantum/tests/unit/test_job_results.py
```diff
@@ -61,8 +61,201 @@ def test_job_for_microsoft_quantum_results_v1_invalid_histogram_returns_raw_resu
...
+                    \"Count\": 1
+                    }
+                ],
+                \"Shots\": [
+                    {
+                    \"Item1\": [
+                        1,
+                        0
+                    ],
+                    \"Item2\": {
```
kikomiss, line 160: nit: could you please also add one more Shots object like below to cover for additional case
```json
"Item3": [ 
   { 
      "Item1": 6, 
      "Item2": true 
   }, 
   { 
       "Item1": 12, 
       "Item2": false 
   } 
]
```

kikomiss:
Could you please add live-tests to the `test_job_results.py` that will cover 3 cases:
- `get_results()` when job output is in `microsoft.quantum-results.v2`, validate that the output of the method is valid
- `get_results_histogram()` when job output is in `microsoft.quantum-results.v2`, validate that the output of the method is valid
- `get_results_shots()` when job output is in `microsoft.quantum-results.v2`, validate that the output of the method is valid

You can take the live-method `test_job_get_results_with_expired_sas_token()` from the same file as an example.

azure-quantum/setup.py
```diff
@@ -22,7 +22,7 @@
 # To make sure that local builds still work without the same environment
 # variables, we'll default to 0.0.1 as a development version.
 
-version = os.environ.get("PYTHON_VERSION", "0.0.1")
+version = os.environ.get("PYTHON_VERSION", "1.2.4")
```
kikomiss, line 25: we don't actually need to set version here as the release pipeline will set the package version automatically

azure-quantum/azure/quantum/job/job.py
```diff
@@ -151,12 +151,193 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
                     return {histogram_values[i]: histogram_values[i + 1] for i in range(0, len(histogram_values), 2)}
                 else: 
                     raise f"\"Histogram\" array has invalid format. Even number of items is expected."
+            elif self.details.output_data_format == "microsoft.quantum-results.v2":
+                if "DataFormat" not in results or results["DataFormat"] != "microsoft.quantum-results.v2":
+                    raise f"\"DataFormat\" was expected to be \"microsoft.quantum-results.v2\" in the Job results for \"{self.details.output_data_format}\" output format."
```
kikomiss, line 156: I know this is just copy-paste issue, but we really should get away from `raise <string>` as it's no longer supported in Python 3. Could you please replace these occurrences with `raise ValueError(<string>)` in this file?

azure-quantum/azure/quantum/job/job.py
```diff
@@ -142,21 +142,212 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
...
+        
+        Raises :class:`azure.quantum.job.JobFailedWithResultsError` if job execution fails, 
+                but failure results could still be retrieved (e.g. for jobs submitted against "microsoft.dft" target).
+
+        :param timeout_secs: Timeout in seconds, defaults to 300
+        :type timeout_secs: float
+        :return: Results dictionary with histogram shots, or raw results if not a json object.
+        :rtype: typing.Any
+        """
+        if self.results is not None:
```
cesarzc, line 198: I see that this check is also present in the `get_results` method, but what is its purpose?

azure-quantum/azure/quantum/job/job.py
```diff
@@ -142,21 +142,212 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
...
+                    # This is handling the BatchResults edge case
+                    shotsArray = []
+                    for i, result in enumerate(results):
+                        if "Shots" not in result:
+                            raise ValueError(f"\"Shots\" array was expected to be in the Job results for result {i} of \"{self.details.output_data_format}\" output format.")
+                        shotsArray.append([self._convert_tuples(shot) for shot in result["Shots"]])
+                    
+                    return shotsArray
+            else:   
+                raise ValueError(f"This method only handles Jobs which are submitted with the V2 output format.")
```
cesarzc, line 330:
```suggestion
                raise ValueError(f"Individual shot results is not supported for jobs using the \"{self.details.output_data_format}\" output format.")
```

cesarzc: This error message provides more details about why this fails.

azure-quantum/azure/quantum/job/job.py
```diff
@@ -142,21 +142,212 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
...
+                return {outcome["Display"]: outcome["Count"] / total_count for outcome in histogram_values}
             
             return results
         except:
             # If errors decoding the data, return the raw payload:
             return payload
 
+    def get_results_histogram(self, timeout_secs: float = DEFAULT_TIMEOUT):
+        """Get job results histogram (in V2 format) by downloading the results blob from the
+        storage container linked via the workspace.
```
cesarzc, line 186:
```suggestion
        """Get job results histogram by downloading the results blob from the
        storage container linked via the workspace.
```

cesarzc: The v2 format is an implementation detail so it's better to avoid it in doc comments.

azure-quantum/azure/quantum/job/job.py
```diff
@@ -142,21 +142,212 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
...
+        
+        Raises :class:`RuntimeError` if job execution fails.
```
cesarzc, line 188: This method also raises `ValueError` in some cases, so it would be good to add this info to the doc comment.

azure-quantum/azure/quantum/job/job.py
```diff
@@ -142,21 +142,212 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
 
             if self.details.output_data_format == "microsoft.quantum-results.v1":
                 if "Histogram" not in results:
-                    raise f"\"Histogram\" array was expected to be in the Job results for \"{self.details.output_data_format}\" output format."
+                    raise ValueError(f"\"Histogram\" array was expected to be in the Job results for \"{self.details.output_data_format}\" output format.")
```
cesarzc, line 145: Since this method now raises a `ValueError`, it would be good to include this infor in its doc comment.

azure-quantum/azure/quantum/job/job.py
```diff
@@ -142,21 +142,212 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
...
+                if len(results) == 1: 
+                    results = results[0]
+                    if "Histogram" not in results:
+                        raise ValueError(f"\"Histogram\" array was expected to be in the Job results for \"{self.details.output_data_format}\" output format.")
+
+                    histogram_values = results["Histogram"]
+                    outcome_keys = self._process_outcome(histogram_values)
+
+                    # Re-mapping object {'Histogram': [{"Outcome": [0], "Display": '[0]', "Count": 500}, {"Outcome": [1], "Display": '[1]', "Count": 500}]} to {'[0]': {"Outcome": [0], "Count": 500}, '[1]': {"Outcome": [1], "Count": 500}}
+                    return {hist_val["Display"]: {"outcome": outcome, "count": hist_val["Count"]} for outcome, hist_val in zip(outcome_keys, histogram_values)}
```
cesarzc, line 242: What would be the recommended way to get the total results count?

azure-quantum/azure/quantum/job/job.py
```diff
@@ -142,21 +142,212 @@ def get_results(self, timeout_secs: float = DEFAULT_TIMEOUT):
...
+                        histogram_values = result["Histogram"]
+                        outcome_keys = self._process_outcome(histogram_values)
+
+                        # Re-mapping object {'Histogram': [{"Outcome": [0], "Display": '[0]', "Count": 500}, {"Outcome": [1], "Display": '[1]', "Count": 500}]} to {'[0]': {"Outcome": [0], "Count": 500}, '[1]': {"Outcome": [1], "Count": 500}}
+                        resultsArray.append({hist_val["Display"]: {"outcome": outcome, "count": hist_val["Count"]} for outcome, hist_val in zip(outcome_keys, histogram_values)})
+
+                    return resultsArray
+
+            else:
+                raise ValueError(f"This method only handles Jobs which are submitted with the V2 output format.")
```
cesarzc, line 260:
```suggestion
                raise ValueError(f"Getting a results histogram with counts instead of probabilities is not a supported feature for jobs using the \"{self.details.output_data_format}\" output format.")
```

cesarzc: Suggestion to hopefully make this error message more user friendly.

azure-quantum/tests/unit/recordings/test_job_get_results_with_expired_sas_token.yaml
```diff
@@ -214,557 +214,21 @@ interactions:
...
       - keep-alive
       content-length:
-      - '1079'
+      - '132'
       content-type:
       - application/json; charset=utf-8
-      transfer-encoding:
-      - chunked
-    status:
-      code: 200
```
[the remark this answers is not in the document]
LawsonGraham: we are regenerating recordings because the service has changed (for V2 output format), so as seen in later commits we push a new suite of recordings

azure-quantum/azure/quantum/job/job.py
LawsonGraham: parse the histogram and sum over the counts... or call get shots and check len of list

cesarzc (approved):
Logic-wise it looks good to me. I tested the changes, and it works well!

@kikomiss could you please double check the changes to the recordings? That part I'm not very familiar with.

azure-quantum/tests/unit/test_job_results.py
```diff
@@ -86,8 +296,8 @@ def decode(): str
         download_data = DowloadDataMock()
         download_data.decode = Mock(return_value=results_as_json_str)
         job.download_data = Mock(return_value=download_data)
-        
-        return job.get_results()
+
+        return job
```
cesarzc, line 300: Could this be the cause of the issue? @kikomiss @LawsonGraham

LawsonGraham: not this - see

def _get_job_results(self, output_data_format, results_as_json_str):
        job = self._mock_job(output_data_format, results_as_json_str)
        
        return job.get_results()
    
    def _get_job_results_histogram(self, output_data_format, results_as_json_str):
        job = self._mock_job(output_data_format, results_as_json_str)

        return job.get_results_histogram()
    
    def _get_job_results_shots(self, output_data_format, results_as_json_str):
        job = self._mock_job(output_data_format, results_as_json_str)
        
        return job.get_results_shots()

kikomiss (approved):
Approved with ignored tests. @kikomiss to fix the tests.

PR #625, opened 2024-07-12
migrate QIR backends to V2 output format by default

LawsonGraham:
upgrading the QIR backends to handle and use V2 output format by default, enabling `get_memory()` to be callable for users. also fixing construction of the `get_counts()` method to use real counts vs. synthetic counts from probablilities


azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -345,6 +347,7 @@ def _format_microsoft_v2_results(self) -> List[Dict[str, Any]]:
         
         status = self.status()
 
+        # TODO: construction of header is quite minimal compared to passthrough flow - evaluate if we want to add more information
```
cesarzc, line 350: What kind of information were you thinking about?

LawsonGraham: I'm not sure - not familiar with what headers are used for really, would be good to discuss

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -52,6 +51,27 @@
...
+    "m",
+    "cx",
+    "cz",
+    "h",
+    "reset",
+    "rx",
+    "ry",
+    "rz",
+    "s",
+    "sdg",
```
cesarzc, line 65: It would be good to double check what QIR gets generated for `sdg` and `tdg`, but I think we should remove them from this list.

cesarzc: Once we check how they get currently generated, we should remove them from all providers as they don't represent neither a provider specific QIR supported gate nor an Azure Quantum gate.

azure-quantum/azure/quantum/qiskit/backends/quantinuum.py
```diff
@@ -99,6 +99,7 @@
         return _get_n_qubits(name)
 
 
+# TODO: do we want to update memory: True here?
```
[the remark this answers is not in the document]
cesarzc: I think we should for consistency across targets.

azure-quantum/azure/quantum/qiskit/job.py
cesarzc: Could you please investigate how is this usually populated in Qiskit?

cesarzc:
The logic changes look good, thanks @LawsonGraham. We should just do our due diligence checking QIR generation for each gateset.

@kikomiss, when you have a chance, could you please take a look at the updated recordings? I reviewed them and they look ok, but it would be good to have a second set of eyes on them.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
LawsonGraham: yep you're right, sdg and tdg are not generated from the transpilation

kikomiss (approved):

PR #630, opened 2024-07-29
qiskit qir by default

LawsonGraham:
migrate our SDK to use QIR backends by default for Qiskit backends


azure-quantum/azure/quantum/qiskit/backends/quantinuum.py
```diff
@@ -30,6 +30,25 @@
...
+    "s",
+    "t",
+    "v",
+    "vdg",
+    "rzz",
+    "measure",
+    "reset",
+]
+
+QUANTINUUM_BASIS_GATES_QIR = [
```
idavis, line 51: Need to review this, v, vdg, zz are not supported for generation.

azure-quantum/azure/quantum/qiskit/backends/rigetti.py
```diff
@@ -23,9 +23,7 @@
     "ry",
     "rz",
     "s",
-    "sdg",
     "t",
-    "tdg",
     "x",
     "y",
```
idavis, line 28: put sdg and tdg back in

azure-quantum/azure/quantum/target/target.py
```diff
@@ -327,6 +327,65 @@ def _get_azure_target_id(self) -> str:
...
+
+        
+        one_qubit_gates = 0
+        multi_qubit_gates = 0
+        measurement_gates = 0
+
+        function_entry_points: list[Function] = []
+        for function in module.functions:
+            if is_entry_point(function):
+                function_entry_points.append(entry_point(module, function.name, 0, 0))
```
cesarzc, line 350:
```suggestion
        function_entry_points: filter(is_entry_point, module.functions)
```

cesarzc: All of this can be replaced by just filtering the list of functions in the module using `is_entry_point`.

The current code would be incorrect because a posible misunderstanding on how `entry_point` works. `entry_point` creates a new entry point function in the module that you pass as argument with the given name. To gather statistics about the gates that are being used you don't really want to modify the module.

azure-quantum/azure/quantum/qiskit/provider.py
```diff
@@ -120,18 +120,31 @@ def backends(self, name=None, **kwargs):
...
+                config = backend.configuration().azure
+                is_default =  config["is_default"]
+                is_passthrough = config["is_passthrough"]
+                name = backend.name()
+
+                # Process only if the backend is marked as default
+                if is_default:
+                    # If the name is not in the dictionary, add the backend
+                    # If the current backend is not passthrough, update it in the dictionary
+                    if name not in filtered_defaults or not is_passthrough:
```
kikomiss, line 136: what is the meaning of `is_passthrough` here and why do we need to update the `filtered_defaults` if we found another backend with the same name that has `is_passthrough = False`?

azure-quantum/tests/unit/test_qiskit.py
```diff
@@ -290,6 +290,7 @@ def test_qiskit_submit_ionq_5_qubit_superposition(self):
         workspace = self.create_workspace()
         provider = AzureQuantumProvider(workspace=workspace)
         self.assertIn("azure-quantum-qiskit", provider._workspace.user_agent)
+        # backend = provider.get_backend("ionq.simulator", input_data_format = "ionq.circuit.v1", output_data_format = "ionq.quantum-results.v1", gateset="qis")
```
kikomiss, line 293: Are we commenting this out just temporarily? Otherwise, if this is not required anymore we can just remove it

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -326,10 +340,14 @@ def _format_microsoft_v2_results(self) -> List[Dict[str, Any]]:
...
 
         results = self._translate_microsoft_v2_results()
-        
+
+        headers = self._get_headers()
         if len(results) != len(entry_point_names):
             raise ValueError("The number of experiment results does not match the number of experiment names")
         
+        if len(results) != len(headers):
+            raise ValueError("The number of experiment results does not match the number of experiment names")
```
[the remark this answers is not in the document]
LawsonGraham: copy paste error - good catch, thank you!

azure-quantum/tests/unit/test_qiskit.py
LawsonGraham: removing

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -312,6 +312,20 @@ def _get_entry_point_names(self):
...
+        headers = self._azure_job.details.metadata
+        # TODO: when multi circuit jobs are possible, confirm metadata field will be a list of metadatas
+        if (not isinstance(headers, list)):
+            headers = [headers]
+        
+        for header in headers:
+            del header['qiskit'] # we throw out the qiskit header as it is implied
+            for key in header.keys():
+                if isinstance(header[key], str) and header[key].startswith('{') and header[key].endswith('}'):
+                    header[key] = json.loads(header[key])
```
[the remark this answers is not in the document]
LawsonGraham: I have changed to trying to dynamically load the header as JSON and reverting if not possible. This should make loading much more flexible.

azure-quantum/azure/quantum/qiskit/provider.py
LawsonGraham: To give a bit more context, now that we have qiskit QIR by default, IonQ passthrough backends are harder to get (as if someone specifies something like `get_backend("ionq.simulator", input_data_format="ionq.circuit.v1")` this will actually still fail (as it matches to both the native and qis gateset now, as qis is no longer the default out of the two). So, this new flag allows for easier fetching of passthrough backends, where we set the default passthroguh and QIR backends to `is_default=True`, and then if we've matched to both backends (the default QIR and passthrough backend), we'll always prefer to use the `is_passthrough=False` backend, aka the QIR backend.

TLDR: is_passthrough is a new flag that basically lets us to still have defaults between passthrough backends, while also maintaining QIR as our main default.

azure-quantum/tests/unit/test_qiskit.py
```diff
@@ -404,6 +406,47 @@ def test_plugins_estimate_cost_qiskit_ionq(self):
...
+        workspace = self.create_workspace()
+        provider = AzureQuantumProvider(workspace=workspace)
+        self.assertIn("azure-quantum-qiskit", provider._workspace.user_agent)
+        provider.get_backend("ionq.simulator", is_passthrough=True)
+        provider.get_backend("ionq.simulator", is_passthrough=True, gateset="native")
+        provider.get_backend("ionq.simulator", input_data_format="ionq.circuit.v1", output_data_format="ionq.quantum-results.v1", gateset="qis")
+        provider.get_backend("ionq.simulator", is_passthrough=False)
+        provider.get_backend("ionq.simulator", is_passthrough=True, input_data_format="ionq.circuit.v1", output_data_format="ionq.quantum-results.v1", gateset="qis")
+        provider.get_backend("ionq.simulator")
+        provider.get_backend("ionq.simulator", input_data_format="qir.v1")
```
kikomiss, line 448: Could you please do `self.assertIsInstance(backend, IonQSimulatorQirBackend)` for all of these backends to make sure the backend returned from `get_backend` is of expected type?

azure-quantum/tests/unit/test_qiskit.py
```diff
@@ -553,6 +596,49 @@ def _test_qiskit_submit_ionq(self, circuit, **kwargs):
...
+        self.assertEqual(qiskit_job._azure_job.details.target, "ionq.simulator")
+        self.assertEqual(qiskit_job._azure_job.details.provider_id, "ionq")
+        self.assertEqual(qiskit_job._azure_job.details.input_data_format, expected_data_format)
+        self.assertEqual(qiskit_job._azure_job.details.output_data_format, MICROSOFT_OUTPUT_DATA_FORMAT_V2)
+        self.assertEqual(qiskit_job._azure_job.details.input_params["shots"], shots)
+        self.assertIn("qiskit", qiskit_job._azure_job.details.metadata)
+        self.assertIn("name", qiskit_job._azure_job.details.metadata)
+        self.assertIn("metadata", qiskit_job._azure_job.details.metadata)
+        # NOTE: this check is no longer required for QIR backends as the `meas_map` metadata field is not needed for QIR
+        # self.assertIn("meas_map", qiskit_job._azure_job.details.metadata)
```
kikomiss, line 619: nit: remove if not needed anymore.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -343,14 +372,23 @@ def __init__(self, name: str, provider: "AzureQuantumProvider", **kwargs):
             "configuration", default_config
         )
         super().__init__(configuration=configuration, provider=provider, **kwargs)
+    
+    def _azure_config(self) -> Dict[str, str]:
+        config = super()._azure_config()
+        config.update(
+            {
+                "is_default": True,
```
cesarzc, line 380: I'm not sure this is correct. Wouldn't all backends derived from `IonQBackend`should keep the default config from their parent class such that `is_default` is set to `False`?

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -392,6 +430,15 @@ def __init__(self, name: str, provider: "AzureQuantumProvider", **kwargs):
...
```
cesarzc, line 438: Similarly to my other comment, I believe this is not correct and the configuration of this backend should not override the one from the parent class.

cesarzc: Same for all the other backends that derive from `IonQBackend`.

azure-quantum/azure/quantum/qiskit/backends/qci.py
```diff
@@ -119,7 +118,6 @@ def __init__(self, name: str, provider: "AzureQuantumProvider", **kwargs):
                 "open_pulse": False,
                 "gates": [{"name": "TODO", "parameters": [], "qasm_def": "TODO"}],
                 "azure": self._azure_config(),
-                "is_default": True,
```
cesarzc, line 122: I think this is not correct either. This should be the default backend for QCI since for this provider QIR is the only option. Changes to this file are not really necessary.

azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -326,10 +346,14 @@ def _format_microsoft_v2_results(self) -> List[Dict[str, Any]]:
         entry_point_names = self._get_entry_point_names()
 
         results = self._translate_microsoft_v2_results()
-        
+
+        headers = self._get_headers()
```
cesarzc, line 350: nit: since the `headers` variable is not used until after the first `if` statement, its assignment can be moved until after it.

azure-quantum/azure/quantum/qiskit/provider.py
```diff
@@ -120,18 +120,37 @@ def backends(self, name=None, **kwargs):
...
-                ),
-                filtered_backends,
-            )
-        ) 
-       # If default backends were found - return them, otherwise return the filtered_backends collection.
-       # The latter case could happen where there's no default backend defined for the specified target.  
-        if len(default_backends) > 0:
+        # The `is_passthrough` flag is used as a more intuitive way to match passthrough backends, and
+        # to allow for us to set default passhtrough backends, while still defaulting to QIR backends
+        # in case we match to both a QIR and a passthrough backend. Without the `is_passthrough`flag
```
cesarzc, line 125:
```suggestion
        # in case we match to both a QIR and a passthrough backend. Without the `is_passthrough` flag 
```

azure-quantum/azure/quantum/qiskit/provider.py
```diff
@@ -120,18 +120,37 @@ def backends(self, name=None, **kwargs):
...
-       # If default backends were found - return them, otherwise return the filtered_backends collection.
-       # The latter case could happen where there's no default backend defined for the specified target.  
-        if len(default_backends) > 0:
+        # The `is_passthrough` flag is used as a more intuitive way to match passthrough backends, and
+        # to allow for us to set default passhtrough backends, while still defaulting to QIR backends
+        # in case we match to both a QIR and a passthrough backend. Without the `is_passthrough`flag 
+        # fetching backends for IonQ require at least 2 input flags to be able to match to a single
+        # backend, those being both `input_data_format` and `gateset` (as there are multiple gateset
+        # versions per passthrough backend) 
+        def filter_non_passthrough_defaults(filtered_backends):
```
cesarzc, line 129: I think I haven't understood the purpose of this function. What made the `_filter_backends` method no longer sufficient to get the filtered list of backends we now expect?

azure-quantum/azure/quantum/target/target.py
```diff
@@ -327,6 +327,52 @@ def _get_azure_target_id(self) -> str:
...
+
+                    # Determine the type of gate based on the counts
+                    if qubit_count == 1 and result_count == 0:
+                        one_qubit_gates += 1
+                    if qubit_count >= 2 and result_count == 0:
+                        multi_qubit_gates += 1
+                    if result_count > 0:
+                        measurement_gates += 1
+
+        return one_qubit_gates, multi_qubit_gates, measurement_gates
```
cesarzc, line 374: The type of this return statement does not seem to match the type of this method's signature.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
LawsonGraham: This change is due to the `is_passthrough` flag... we want to specify a default passthrough and a default QIR backend, and then in filtration of the backends, we can handle whether to use the QIR or passthrough default... this change was made because if not, then selection of ionq backends for example would require specification of both input format and gateset. With the `is_passthrough` flag, we can still maintain the same defaults for passthrough backends, while also having defaults for QIR backends. In filtration we then prefer QIR defaults > passthrough defaults

azure-quantum/azure/quantum/qiskit/provider.py
LawsonGraham: `_filter_backends`'s default backend filtration logic previously only considered `is_default` flag... now, given we have both passthrough and QIR defaults, we must consider the `is_passthrough` flag, and if there is no `is_passthrough` flag, we prefer QIR defaults over passthrough defaults. So, the default selection / filtration logic got a bit more complicated

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -46,6 +45,24 @@
     "IonQForteNativeBackend",
 ]
 
+QIR_BASIS_GATES = [
```
cesarzc, line 48: The only missing gate is `reset`.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
```diff
@@ -68,6 +85,9 @@ def _default_options(cls) -> Options:
             },
             targetCapability="BasicExecution",
             )
+    
+    def gateset(self):
```
cesarzc, line 89: For all QIR backends the `gateset` configuration property should be `qis`. I think this can be done the same way that `basis_gates` is done, which is by creating a private `_gateset` method in the `AzureQirBackend` class and using it when populating the configuration.

azure-quantum/azure/quantum/qiskit/provider.py
cesarzc: After thinking about it, I believe having extra logic to being able to create a pass-through backend is not worth it. The reason is that functionally, a pass-through backend does not offer additional features than a QIR backend, with the exception of IonQ's native backends that provide a specialized gateset (in which case a user would filter based on the `gateset` configuration property).

We should consider using pass-through backends an advanced feature, so it's ok to have to specify multiple filtering properties is needed (like in the case of IonQ).

azure-quantum/azure/quantum/target/target.py
```diff
@@ -327,6 +327,52 @@ def _get_azure_target_id(self) -> str:
     def _get_azure_provider_id(self) -> str:
         return self.provider_id
 
+    def _qir_module_to_gates(self, qir_module) -> List[int]:
```
cesarzc, line 330: Regarding this method, there are a couple of things that could be changed:
- It should probably be a class method.
- Consider returning a data class object. Returning a list makes it difficult to figure out what is the meaning of a number at a particular index is, and it also makes its use more error prone.
- Consider changing to name to something that reflects more accurately what is calculates like `_calculate_qir_module_gate_stats`.

cesarzc:
@LawsonGraham, could you update your development branch with the latest changes from `main`? There is one change that was recently merge that will probably have a conflict with your changes.

@kikomiss, could you please take a look at the recording files to check that they are correct? Also, I believe the security alerts can be dismissed given that they're just triggering because the recordings we get back contain the word "md5", but I wanted to double check with you in case I'm missing something.

azure-quantum/azure/quantum/qiskit/backends/ionq.py
LawsonGraham: Ok, just to be very clear as I remember we had some confusion here... I will use this basis gate set:

```
QIR_BASIS_GATES = [
    "x",
    "y",
    "z",
    "rx",
    "ry",
    "rz",
    "h",
    "swap",
    "cx",
    "cz",
    "reset",
    "s",
    "sdg",
    "t",
    "tdg",
    "measure",
]
```

for IonQ, Rigetti, Quantinuum. What about QCI?

For context Quantinuum currently has a different set of basis gates from ionq and rigetti. please confirm @cesarzc @idavis

azure-quantum/azure/quantum/qiskit/backends/ionq.py
LawsonGraham: the notion of `gateset` as a parameter only really exists in IonQ's backends... because of this I added the gateset method in IonQ's QIR backend base. in fact, the gate set field is not actually even used for anything in the QIR IonQ backend... I only added it to have consistency in flags across IonQ backends, and to keep a command like `get_backend("ionq.simulator", gateset="qis')` non-breaking while still pointing to a QIR backend. For now, I will leave it as is. Let me know if we prefer to move it, or remove it, or leave.

azure-quantum/azure/quantum/target/target.py
LawsonGraham: good suggestion!

azure-quantum/azure/quantum/qiskit/backends/ionq.py
LawsonGraham: quantinuum's basis gates: 

```
QUANTINUUM_BASIS_GATES = [
    "x",
    "y",
    "z",
    "rx",
    "ry",
    "rz",
    "h",
    "cx",
    "cz",
    "s",
    "sdg",
    "t",
    "tdg",
    "v",
    "vdg",
    "zz",
    "measure",
    "reset",
]
```

LawsonGraham: when using these default QIR gates for Quantinuum circuits like `stswap` fail... (as `swap`is not a basis gate for quantinuum)

I recommend we have some default QIR basis gates to be used by rigetti, ionq, (maybe qci?), and then have a separate set for quantinuum.

cesarzc: Just double checked this and I agree. IonQ, Rigetti and QCI can share the same gateset, and Quantinuum should have its own gateset that does not contain swap. Both basis gates list should be the same with the exception of swap.

cesarzc: Also confirming that the `QIR_BASIS_GATES` that you listed is correct.

LawsonGraham: sounds good - added this logic

azure-quantum/azure/quantum/qiskit/backends/quantinuum.py
```diff
@@ -94,9 +94,20 @@ def _azure_config(self) -> Dict[str, str]:
             }
         )
         return config
+    
+    def _basis_gates(self) -> List[str]:
+        # We use all normal QIR basis gates, with the exception of `swap`
```
cesarzc, line 99: Even though the difference between the Quantinuum basis gates and the rest of the providers is just the swap gate, it would be easier to see the basis gates if it was an explicit list.

cesarzc:
@LawsonGraham, the logic looks good!

Once the CI pipeline errors are addressed and @kikomiss confirms whether the changes to recordings are correct, I'll approve.

azure-quantum/azure/quantum/qiskit/provider.py
```diff
@@ -129,9 +129,10 @@ def backends(self, name=None, **kwargs):
                 filtered_backends,
             )
         ) 
-       # If default backends were found - return them, otherwise return the filtered_backends collection.
-       # The latter case could happen where there's no default backend defined for the specified target.  
-        if len(default_backends) > 0:
+
+        # If default backends were found - return them, otherwise return the filtered_backends collection.
+        # The latter case could happen where there's no default backend defined for the specified target.  
+        if len(default_backends) > 0:
```
idavis, line 135: nit: remove trailing whitespace

kikomiss (approved):

cesarzc (approved):

PR #635, opened 2024-08-08
qubit count update

ZachMassa:
Updating the H2-1 target to 52 qubits to match current performance specifications.


kikomiss:
@ZachMassa LGTM, I think we also would need to update number of qubits for h2-1 here
https://github.com/microsoft/azure-quantum-python/blob/778a0896c95d28fcc8f61e0d735ac2e0bdac81ea/azure-quantum/azure/quantum/qiskit/backends/quantinuum.py#L62

kikomiss:
Sorry, just found why the unit-tests are failing. We're also going to need to update this test:
https://github.com/microsoft/azure-quantum-python/blob/7f96bb502fdfecef14217dac0cf4431d89e349c4/azure-quantum/tests/unit/test_qiskit.py#L1176
it assumes that `h2-1` uses 32 qubits

xinyi-joffre:
@ZachMassa, following up on this PR to see if you still wanted to merge this? I think there's the remaining failing unit test that @kikomiss pointed out, and then we can merge.

ZachMassa:
> @ZachMassa, following up on this PR to see if you still wanted to merge this? I think there's the remaining failing unit test that @kikomiss pointed out, and then we can merge.

I am fine with merging it. It seems like the issue mentioned above is that the unit test assumed H2 has a qubit count of 32 instead of 56.

xinyi-joffre:
@ZachMassa , can you make the fix for the unit test, since these changes are on your fork?

ZachMassa:
> @ZachMassa , can you make the fix for the unit test, since these changes are on your fork? Otherwise, the pipelines won't pass in order for your PR to merge.

I believe that these test should be updated to reflect the new qubit count now.

kikomiss (approved):

xinyi-joffre:
Thank you for the update, @ZachMassa ! It's been merged!

PR #641, opened 2024-09-25
Use qsharp package to generate QIR

idavis:
This change replaces the usage of `qiskit-qir` with the `qsharp` package. Of note:
- `targetCapablities` is deprecated where possible and replaced with `TargetProfile`. Since `qsharp` is an optional dependency when not using Qiskit, the type is forward declared with a `str` union option.
- New `target_profile` parameter is set to replace `targetCapablities` where possible.
- Job submissions should be sending `target_profile` for metadata.
- Logging of pre-post translation to QIR is changed.
- Basis gates are unified based on the `qsharp` QIR generation intrinsics. All changes to gateset are handled with decompositions with the exception of `barrier` which is only sent for a single target. There had been changes to the QIR job submission gate sets that broke this behavior.
- `azure/quantum/qiskit/backends/quantinuum.py` had the default profile/caps set to basic execution. This PR fixes the profile setting for this backend.
- While the `qsharp` package can perform local resource estimation, the existing service submission is left intact with the new QIR generation code paths. PyQIR is used to implement output recording calls which are always part of `qsharp` code generation, but aren't handled by the service.
- Register order no longer has to be inverted client side. The `qsharp` QIR generation correctly sets endianness of output recording to match Qiskit's semantics. This does mean that accessing data from jobs submitted with earlier versions will have incorrect result rendering.


azure-quantum/azure/quantum/qiskit/job.py
```diff
@@ -212,15 +212,20 @@ def _format_ionq_results(self, sampler_seed=None):
...
     def _qir_to_qiskit_bitstring(obj):
-        """Convert the data structure from Azure into the "schema" used by Qiskit """
+        """Convert the data structure from Azure into the "schema" used by Qiskit"""
         if isinstance(obj, str) and not re.match(r"[\d\s]+$", obj):
             obj = ast.literal_eval(obj)
 
         if isinstance(obj, tuple):
             # the outermost implied container is a tuple, and each item is
             # associated with a classical register. Azure and Qiskit order the
             # registers in opposite directions, so reverse here to match.
```
kikomiss, line 222: nit: as far as I understood, the registers are no longer in opposite direction, so this comment is no longer valid?

kikomiss (approved):

kikomiss (approved):

kikomiss (approved):

PR #647, opened 2024-10-18
Add batching support to sdk

adam-grofe:
This PR is to lay the foundation for the batching feature where we provide convenience functions for users to provide a list of xyz files and then the SDK converts these to QcSchema to be uploaded to the service. The QcSchema are uploaded as individual blobs to the storage account container used for the job. This PR takes advantage of the container URI that is already be used to save the output to. 

Tests were defined for the functions that generate QcSchema using pytest and pytest-regressions. End-to-end testing was performed manually by submitting a single job batch to canary for each job type.


azure-quantum/azure/quantum/target/microsoft/elements/dft/target.py
```diff
@@ -73,15 +75,153 @@ def submit(self,
...
+        """
+
+        qcshema_objects = []
+        for file in input_data:
+            file_path = Path(file)
+            if not file_path.exists():
+                raise FileNotFoundError(f"File {file} does not exist.")
+            
+            file_data = file_path.read_text()
+            mol = self._xyz_to_qcschema_mol(file_data)
```
kikomiss, line 128: Should we handle the case where provided input files are already a `QCSchema`? In such case, we will just add them to `qcschema_objects` without transformation. We could naively assume format from the file extension. 
There could also be a mix of `.xyz` and `.qcschema` files provided to the job, so handling them in this loop still makes sense.

kikomiss (approved):

azure-quantum/azure/quantum/target/microsoft/elements/dft/target.py
adam-grofe: Done

adam-grofe: I used `.json` as the file extension for qcschema. I also added a warning in case they provide both input parameters and a qcschema. It will use the parameters in the qcschema rather than modifying it with the input parameters.

kikomiss (approved):

PR #649, opened 2024-11-08
Disable QCI tests

04diiguyi:


xinyi-joffre (approved):

azure-quantum/tests/unit/test_qiskit.py
```diff
@@ -1634,8 +1634,8 @@ def test_qiskit_get_rigetti_qpu_targets(self):
         self.assertEqual("qir.v1", config.azure["input_data_format"])
         self.assertEqual(MICROSOFT_OUTPUT_DATA_FORMAT_V2, backend._get_output_data_format())
 
-    @pytest.mark.qci
-    @pytest.mark.live_test
+    #@pytest.mark.qci
```
kikomiss, line 1637: very-nit: could we use 
```python
@pytest.mark.skip("Skipping tests against QCI's unavailable targets")
```
for traceability why we're excluding these tests.

kikomiss (approved):

PR #655, opened 2025-01-17
fix: install pytest_regressions in release pipeline

kikomiss:


almullay (approved):
Approved, changes look good.

PR #669, opened 2025-03-21
Remove additional resource estimator files

xinyi-joffre:
Remove files leftover from resource estimator


azure-quantum/azure/quantum/target/params.py
almullay, line 1: After removing `azure.quantum.target.params.py` in the Resource Estimator deprecation effort, we're seeing CI failures across multiple test files:

```
ModuleNotFoundError: No module named 'azure.quantum.target.params'
```

This is affecting:
- test_authentication.py
- test_cirq.py
- test_job_results.py
- test_ionq.py

Despite deprecating the Resource Estimator target, it appears other parts of the codebase still rely on the `InputParams` class from this module. Should we restore it, or add a minimal version of the file to unblock CI?

04diiguyi (approved):

almullay (approved):

PR #676, opened 2025-06-10
Relax upper bound for numpy

xinyi-joffre:


vtomole:
Hey @xinyi-joffre , @kikomiss  can we please merge this in?

PR #680, opened 2025-08-28
Deprecate DFT target

xinyi-joffre:
The private preview DFT target in Azure Quantum is being deprecated and moved to its new home in Microsoft Discovery.

Removing code and tests for DFT.


kikomiss (commented):
I think we could also remove `pytest-regressions` package that gets installed in `.ado/ci.yml` and in the `environment.yml`. It was added as part of this [DFT commit](https://github.com/microsoft/azure-quantum-python/commit/20dc5cbdf1614107c7aa2dbfbb83a397c3f99ee4#diff-cc57b598d819e5277e40e331f6063d939cdc3dd0dc455eee5f4753105f99058a). We don't use this framework anywhere else.

almullay (approved):

PR #738, opened 2026-03-30
feat: add rigetti target cepheus-1-108q, remove ankaa-3

jselig-rigetti:


rigidit:
Thanks for this PR, @jselig-rigetti!

We need to phase this transition to ensure a safe migration path, so we aren't quite ready to fully deprecate the old target just yet. We've merged PR #740 with the addition of the new target first.

If you'd like to own the deprecation portion, feel free to rebase this PR on the latest main so it only contains the removal logic, and we'll merge it later. Otherwise, we can close this PR and take care of the deprecation ourselves in the future. Just let us know your preference!

ScottCarda-MS:
The ankaa-3 model was phased out with this PR: #741. I believe this, combined with the earlier #740,  completes the work that is present in this PR, so I'll close it out.

PR #743, opened 2026-04-30
Add new Pasqal targets

Fenkiou:
Add new emulator targets and fresnel can1 qpu target.

This is a follow up of https://github.com/MicrosoftDocs/quantum-docs/pull/690

### DIFF
--- a/azure-quantum/azure/quantum/target/pasqal/target.py
+++ b/azure-quantum/azure/quantum/target/pasqal/target.py
@@ -45,17 +45,17 @@ class PasqalTarget(str, Enum):
     def simulators() -> List[str]:
         """Returns a list of simulator targets"""
         return [
-            PasqalTarget.SIM_EMU_TN.value
-            PasqalTarget.SIM_EMU_SV.value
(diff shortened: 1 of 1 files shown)


azure-quantum/azure/quantum/target/pasqal/target.py
```diff
@@ -26,28 +26,52 @@ class PasqalTarget(str, Enum):
 
     SIM_EMU_TN = "pasqal.sim.emu-tn"
     """pasqal.sim.emu-tn target"""
```
Fenkiou, line 28: We'd like to deprecate this one and already removed it from the doc there: https://learn.microsoft.com/en-us/azure/quantum/provider-pasqal.

Can we just remove it from here? Not sure how you handle this kind of breaking changes

rigidit: Please create separate PR to deprecate the target.

rigidit (approved):

rigidit (approved):

ScottCarda-MS (approved):

PR #758, opened 2026-07-13
Add support for job delete operation

v-elegacheva:
adds support for job deletion

### DIFF
--- a/azure-quantum/azure/quantum/cirq/job.py
+++ b/azure-quantum/azure/quantum/cirq/job.py
@@ -150,7 +150,7 @@ class Job:
 
     def delete(self):
         """Delete the given job."""
-        self._azure_job.workspace.cancel_job(self._azure_job)
+        self._azure_job.workspace.delete_job(self._azure_job)
 
     def __str__(self) -> str:
         return f"azure.quantum.cirq.Job(job_id={self.job_id()})"
--- a/azure-quantum/azure/quantum/workspace.py
+++ b/azure-quantum/azure/quantum/workspace.py
@@ -465,12 +465,14 @@ class Workspace:
         return Job(self, details)
 
     def delete_job(self, job: Job) -> None:
-        """
-        Deletes a job.
+        """Deletes a job.
+
+        Note: For service API versions before 2025-12-01-preview, the underlying
+        `jobs.delete` operation acts as a cancellation request rather than a hard delete.
 
         :param job:
             Job to delete.
-            """
+        """
         client = self._get_jobs_client()
         client.delete(
             self.subscription_id,
--- a/azure-quantum/azure/quantum/workspace.py
+++ b/azure-quantum/azure/quantum/workspace.py
@@ -464,6 +464,20 @@ class Workspace:
             job.id)
         return Job(self, details)
 
+    def delete_job(self, job: Job) -> None:
+        """
+        Deletes a job.
+
+        :param job:
+            Job to delete.
+            """
+        client = self._get_jobs_client()
+        client.delete(
+            self.subscription_id,
+            self.resource_group,
+            self.name,
+            job.details.id)
+
     def get_job(self, job_id: str) -> Job:
         """
         Returns the job corresponding to the given id.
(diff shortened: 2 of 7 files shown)


azure-quantum/azure/quantum/cirq/job.py
```diff
@@ -150,7 +150,7 @@ def cancel(self):
 
     def delete(self):
         """Delete the given job."""
-        self._azure_job.workspace.cancel_job(self._azure_job)
+        self._azure_job.workspace.delete_job(self._azure_job)
```
rigidit, line 153: would you like to update also 
azure-quantum\azure\quantum\cirq\targets\ionq.py?
maybe other places where delete operation would be appropriate?

azure-quantum/tests/test_workspace.py
```diff
@@ -384,6 +384,34 @@ def test_workspace_cancel_job_success():
...
+
+    ws._client.services.jobs._store.append(details)
+
+    job = Job(ws, details)
+
+    ws.delete_job(job)
+
+    stored_job = ws.get_job(job_id)
+
+    assert stored_job.details.status == "Cancelled"
```
rigidit: so when we delete the job the status should be "Cancelled"? what's the difference with cancel operation then?

azure-quantum/azure/quantum/cirq/job.py
v-elegacheva: Thank you for the catch! I I updated the Ionq and qiskit to use the delete operation and pushed the changes. I also reran the E2E and verified the delete flow still works successfully

azure-quantum/tests/test_workspace.py
v-elegacheva: That is a good question... From my understanding, the "Cancelled" status is coming from the current WorkspaceMock implementation instead of actual service behavior.
During the E2E validation, a deleted job was no longer retrievable after delete_job() succeeded. It did not transition to a "deleted" status, it just disappeared from the service.
I could update the mock to use a "Deleted" status instead, but I am not sure if "Deleted" would be a valid job status in the real service. My understanding is that once a job is deleted, it is removed rather than entering a new status. How would you prefer I proceed?

rigidit: see comment above from Copilot: "Consider updating the mock delete implementation to remove the job from the store and changing this test to assert ws.get_job(job_id) raises (e.g., KeyError in the mock, ResourceNotFoundError in real client), so the test actually covers delete semantics."

azure-quantum/azure/quantum/workspace.py
```diff
@@ -464,6 +464,22 @@ def cancel_job(self, job: Job) -> Job:
             job.id)
         return Job(self, details)
 
+    def delete_job(self, job: Job) -> None:
+        """Deletes a job.
+
+        Note: For service API versions before 2025-12-01-preview, the underlying
```
rigidit: I think [_client ](https://github.com/microsoft/azure-quantum-python/tree/main/azure-quantum/azure/quantum/_client) directory contains autogenerated client based on one of the latest api versions >= 2025-12-01-preview supported by our typespecs, so we can probably just remove comment that it was cancellation request before that version.

v-elegacheva: Sounds good! I will remove comment

v-elegacheva:
@microsoft-github-policy-service agree

v-elegacheva:
@microsoft-github-policy-service agree company="Akvelon"

rigidit (approved):

azure-quantum/tests/test_workspace.py
```diff
@@ -384,6 +385,59 @@
...
+        resource_group=RESOURCE_GROUP,
+        name=WORKSPACE
+    )
+
+    job_id = "test-delete-success"
+
+    details = JobDetails(
+        id=job_id,
+        name=job_id,
+        container_uri="http://example.com/container",
```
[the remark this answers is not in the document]
v-elegacheva: updated the test URL to use https

rigidit (approved):

PR #762, opened 2026-07-27
Support full-result format in Pasqal Result

Fenkiou:
Pasqal jobs previously uploaded only a subset of the result (raw counter dict). Now that full results will be uploaded, the output may be wrapped as `{"counter": {...}, ...}`. Unwrap "counter" when present, falling back to the raw dict for backward compatibility.

Related changes:
- https://github.com/MicrosoftDocs/quantum-docs/pull/698
- https://github.com/pasqal-io/pulser-azure/pull/13

### DIFF
--- a/azure-quantum/azure/quantum/target/pasqal/result.py
+++ b/azure-quantum/azure/quantum/target/pasqal/result.py
@@ -10,7 +10,7 @@ __all__ = [
 ]
 
 import json
-from typing import Union, Dict, List, TypeVar, cast
+from typing import Any, Union, Dict, List, TypeVar, cast
 
 from ...job import Job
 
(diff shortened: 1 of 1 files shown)


azure-quantum/azure/quantum/target/pasqal/result.py
```diff
@@ -44,7 +44,10 @@ def __init__(self, job: Job) -> None:
                 f"(status: {job.details.status}."
                 f"error: {job.details.error_data})"
             )
-        self.data = cast(Dict[str, int], json.loads(job.download_data(job.details.output_data_uri)))
```
Fenkiou, line 47: can be merged as changes are backward compatible

ScottCarda-MS (approved):
This looks good! Thanks for updating to support the full format!

PR #764, opened 2026-08-04
Add update job operation to support updating name, priority and tags

v-elegacheva:
## Summary
Adds support for updating a submitted job's name, priority, and/or tags to the Python SDK, implementing the SDK side of the merged Data Plane API Spec. Only the fields you provide are sent, and the updated job is returned. Note: `tags` replaces existing tags rather than appending.

## What Changed
- Added `Workspace.update_job(job, *, name=None, priority=None, tags=None)`.
- Added the `Job.update(...)` instance method, which updates the job in place.
- Requires at least one field (`name`, `priority`, or `tags`); raises `ValueError` otherwise.
- Exposed the `Priority` enum (`Standard` / `High`) publicly.
- Added unit tests and mock client support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -408,6 +408,3 @@ azure-quantum/build
 !azure-quantum/requirements*.txt
 [.][v]env/
 version.py
-
-# Local scratch/manual test file (not part of the package)
-azure-quantum/test_targets.py
--- a/.gitignore
+++ b/.gitignore
@@ -408,3 +408,6 @@ azure-quantum/build
 !azure-quantum/requirements*.txt
 [.][v]env/
 version.py
+
+# Local scratch/manual test file (not part of the package)
+azure-quantum/test_targets.py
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -408,3 +408,6 @@ azure-quantum/build
 !azure-quantum/requirements*.txt
 [.][v]env/
 version.py
+
+# Local scratch/manual test file (not part of the package)
+azure-quantum/test_targets.py
```
ScottCarda-MS: Was this accidentally left in? This isn't a file that gets generated in this repo or used by us. If it is local to your own setup and you don't want to have to deal with it for commits, you could add it to your own .git/info/exclude file so that it doesn't show up in the repo at all.

v-elegacheva: All fixed! Removed the test_targets.py entry from .gitignore and added it to my local exclude instead. Thank you :)

ScottCarda-MS (approved):
Thank you for the contribution! Looks good to merge!

PR #768, opened 2026-08-11
Quantum: Add Job.list_attachments()

v-elegacheva:
## Summary 

Adds a new `Job.list_attachments()` method to the Azure Quantum Python SDK that lists every blob in a job's Azure Storage output container, returning lightweight metadata (name, size, last modified) for each. This gives Python users a programmatic way to discover a job's associated files, without downloading them, complementing the existing `upload_attachment` / `download_attachment` methods.

## Changes 
- base_job.py- Added `list_attachments() -> list[BlobProperties]`. It resolves the job's linked storage container (from `self.details.container_uri`, falling back to `workspace.get_container_uri(job_id=self.id)` ), opens a `ContainerClient`, and returns `list(container_client.list_blobs()). Returning `BlobProperties` keeps the API consisten with other types listing APIs in the package rather than returning ad-hoc dicts.
- `test_job_attachments.py` (new)- unit tests

## Testing 
`testing_job_attachments.py` - 2 tests, all passing: 
- `test_list_attachments_returns_container_blobs`: returns the container's blob as-is.
- `test_list_attachments_uses_workspace_container_when_unset`: falls back to `workspace.get_container_uri(job_id= ...)` when the job has no linked container URI.
- Manual E2E test against a real Azure Quantum workspace

### DIFF
--- a/azure-quantum/azure/quantum/job/base_job.py
+++ b/azure-quantum/azure/quantum/job/base_job.py
@@ -10,7 +10,7 @@ from enum import Enum
 from datetime import datetime, timezone, timedelta
 from urllib.parse import urlparse, parse_qs
 from typing import Any, Dict, Optional, TYPE_CHECKING
-from azure.storage.blob import BlobClient
+from azure.storage.blob import BlobClient, BlobProperties
 
 from azure.quantum.storage import upload_blob, download_blob, download_blob_properties, ContainerClient
 from azure.quantum._client.models import JobDetails
@@ -388,6 +388,25 @@ class BaseJob(WorkspaceItem):
         blob_client = container_client.get_blob_client(name)
         response = blob_client.download_blob().readall()
         return response
+
+
+    def list_attachments(self) -> list[BlobProperties]:
+        """ Lists the attachments in the job's linked storage container. Attachments are blobs of
+            data created as part of the Job's execution, or they can be uploaded directly from Python
+            using the upload_attachment method.
+
+        :return: List of blobs in the job's linked storage container.
+        :rtype: list[~azure.storage.blob.BlobProperties]
+        """
+
+        # Use the job's linked storage container.
+        if self._details.container_uri is None:
+            container_uri = self.workspace.get_container_uri(job_id=self.id)
+        else:
+            container_uri = self._details.container_uri
+
+        container_client = ContainerClient.from_container_url(container_uri)
+        return list(container_client.list_blobs())
 
 
     def _get_blob_uri_with_sas_token(self, blob_uri: str) -> str:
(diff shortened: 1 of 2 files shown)


azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -390,6 +390,39 @@ def download_attachment(
         return response
 
 
+    def list_attachments(
+        self,
+        container_uri: Optional[str] = None
```
rigidit: let's not allow to define container uri - we should use linked storage account

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -390,6 +390,39 @@ def download_attachment(
...
+                container_uri = self.workspace.get_container_uri(job_id=self.id)
+            else:
+                container_uri = self._details.container_uri
+
+        container_client = ContainerClient.from_container_url(container_uri)
+        return [
+            {
+                "name": blob.name,
+                "size": blob.size,
+                "lastModified": blob.last_modified.isoformat() if blob.last_modified else None,
```
rigidit: in what format we have blob.last_modified? wondering if we really need to convert to iso format here

rigidit: Public Python API surface elsewhere in this package is snake_case — last_modified would be more consistent

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -390,6 +390,39 @@ def download_attachment(
...
+            if self._details.container_uri is None:
+                container_uri = self.workspace.get_container_uri(job_id=self.id)
+            else:
+                container_uri = self._details.container_uri
+
+        container_client = ContainerClient.from_container_url(container_uri)
+        return [
+            {
+                "name": blob.name,
+                "size": blob.size,
```
rigidit: can we document that size in bytes?

rigidit (changes requested):

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -390,6 +390,39 @@ def download_attachment(
...
+                "lastModified": blob.last_modified.isoformat() if blob.last_modified else None,
+            }
+            for blob in container_client.list_blobs()
+        ]
```
rigidit: actually can we just return list[BlobProperties]? 
other listing APIs returns typed models, so it would be better to not return dicts but something typed, and because there is already download_blob_properties method that returns BlobProperties I think we can return list of that model here too

rigidit (approved):

ScottCarda-MS (approved):

PR #769, opened 2026-08-13
Fix typo in signal-processing.ipynb

schrodinteq:
Fix typo in signal-processing.ipynb

`# Map SumPy phases into NumPy`
-> `# Map SymPy phases into NumPy`


ScottCarda-MS (approved):
Thanks for the fix!

PR #770, opened 2026-08-17
Update network isolation policy

kaverkiev:
Update network isolation policy. 
Change Permissive to Preferred. Add GitHub and ESRP


ScottCarda-MS (approved):
Thanks for looking into this!

PR #772, opened 2026-09-11
Update autogenerated client

pavelbraginskiy:
`_client` built from a newer version of the REST API specs.
Matches the client in the Azure CLI, updated here: Azure/azure-cli-extensions/pull/10285

The latest version of the code-generation tools used for the REST API depend on Node >= 22, so I bumped the version mentioned in the readme.

PR #773, opened 2026-09-18
Fix job attachment 401s for submitted jobs

v-elegacheva:
## Summary

Fixes job attachment operations failing after submission when the stored container URI is unsigned, expired, or lacks the permission required by the operation.

`upload_attachment`, `download_attachment`, and `list_attachments` now:

- preserve an explicitly supplied method-level container URI;
- reuse a job-level SAS only when it is unexpired and grants the required permission (`w`, `r`, or `l`);
- otherwise request a fresh SAS for the job's actual container name; and
- reject a refreshed URI when its storage hostname differs from the stored job URI.

Custom container names remain supported. The legacy `Workspace(storage=...)` connection-string path is intentionally outside this PR's scope; its deprecation will be tracked separately.

## Testing

- Focused local validation: `31 passed` (`test_job_attachments.py` and `test_workspace.py`).
- Azure Quantum Python CI: [build 183619](https://dev.azure.com/ms-azurequantum/AzureQuantum/_build/results?buildId=183619) passed.
- The relevant DPv2 attachment E2E in [build 183473](https://dev.azure.com/ms-azurequantum/AzureQuantum/_build/results?buildId=183473) passed upload, list, and download (`34 passed, 2 skipped` in the DPv2 SDK job). The broader live build had unrelated CLI and third-party failures.

## Related

The generated job-update endpoint correction is intentionally left to #772, which regenerated the complete client with the corrected `/jobs/{jobId}` route.


azure-quantum/azure/quantum/_client/operations/_operations.py
```diff
@@ -130,7 +130,7 @@ def build_services_jobs_update_request(
     accept = _headers.pop("Accept", "application/json")
 
     # Construct URL
-    _url = "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Quantum/workspaces/{workspaceName}/jobUpdateOptions/{jobId}"
+    _url = "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Quantum/workspaces/{workspaceName}/jobs/{jobId}"
```
rigidit: we should update all files in autogenerated part instead, review https://github.com/microsoft/azure-quantum-python/pull/772

v-elegacheva: Addressed in recent commit. I removed the hand-edited generated client change and its route test from this PR.  pr #772 owns the full client regeneration and already generates the corrected `/jobs/{jobId}` route;  this pr is now attachment-only and was validated both independently and merged with #772.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -344,12 +344,13 @@ def upload_attachment(
...
+        # unsigned, so always fetch a fresh SAS-signed URI instead of reusing it.
         if container_uri is None:
-            if self._details.container_uri is None:
-                container_uri = self.workspace.get_container_uri(job_id=self.id)
-            else:
-                container_uri = self._details.container_uri
+            container_uri = self.workspace.get_container_uri(
+                job_id=self.id,
+                container_name=self.container_name,
+            )
```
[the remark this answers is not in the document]
v-elegacheva: Addressed in most recent commit. I centralized URI selection for all three attachment methods:
- an explicetely supplied method-level uri is used unchanged
- a job-level sas uri is reused only when it is unexpired and has the permission required for the operation
- otherwise he sdk refreshes the sas using the actual container name and verifies that the refreshed hostname matches the job uri's hostname
- a hostname mismatch raises a clear error before any sotrage operation, so the sdk cannot silently redirect to a same-named container in another accnt

I added focused coverage for valid, expired, insufficient-permission, custom-container, and hostname-mismatch cases.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +391,51 @@ def list_attachments(self) -> list[BlobProperties]:
...
+                    token_expire_time = token_expire_time.replace(tzinfo=timezone.utc)
+                if datetime.now(tz=timezone.utc) < token_expire_time - timedelta(minutes=5):
+                    return container_uri
+            except ValueError:
+                pass
+
+        refreshed_container_uri = self.workspace.get_container_uri(
+            job_id=self.id,
+            container_name=self.container_name,
+        )
```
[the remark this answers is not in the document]
v-elegacheva: Addressed in recent commit. Connection-string workspaces now generate a container-level SAS using `ContainerSasPermissions` with read/add/create/list permissions. 
I also added a focused test that inspects the generated permission object and verifies list access is included.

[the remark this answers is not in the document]
v-elegacheva: Follow-up: after confirming the backend direction with Zulfat, commit a9c1506 removes the connection-string-specific storage.py change to keep #773 scoped to workspace-linked storage. Formal deprecation of Workspace(storage=...) will be tracked separately. The core permission-aware SAS refresh and hostname validation remain unchanged, and the focused tests plus CI pass.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +391,51 @@ def list_attachments(self) -> list[BlobProperties]:
...
-            container_uri = self.workspace.get_container_uri(job_id=self.id)
-        else:
-            container_uri = self._details.container_uri
+        container_uri = self._get_attachment_container_uri(required_permission="l")
 
         container_client = ContainerClient.from_container_url(container_uri)
         return list(container_client.list_blobs())
 
 
+    def _get_attachment_container_uri(self, required_permission: str) -> str:
```
rigidit: required_permission: str is stringly-typed. Use Literal["r", "w", "l"] or a small enum, and give it a return type

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +391,51 @@ def list_attachments(self) -> list[BlobProperties]:
...
+            try:
+                token_expire_time = datetime.fromisoformat(
+                    token_expire_query_param[0].replace("Z", "+00:00")
+                )
+                if token_expire_time.tzinfo is None:
+                    token_expire_time = token_expire_time.replace(tzinfo=timezone.utc)
+                if datetime.now(tz=timezone.utc) < token_expire_time - timedelta(minutes=5):
+                    return container_uri
+            except ValueError:
+                pass
```
rigidit: except ValueError: pass swallows a malformed se silently. logger.debug(...) it — this repo already uses logger liberally and a silent fallback here is hard to diagnose.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +391,51 @@ def list_attachments(self) -> list[BlobProperties]:
...
+        ):
+            try:
+                token_expire_time = datetime.fromisoformat(
+                    token_expire_query_param[0].replace("Z", "+00:00")
+                )
+                if token_expire_time.tzinfo is None:
+                    token_expire_time = token_expire_time.replace(tzinfo=timezone.utc)
+                if datetime.now(tz=timezone.utc) < token_expire_time - timedelta(minutes=5):
+                    return container_uri
+            except ValueError:
```
rigidit: Line 408 - 422 these actually looks like dead branch because self._details.container_uri will return not sas uri, isn't it?

azure-quantum/tests/test_job_attachments.py
```diff
@@ -34,21 +50,218 @@ def test_list_attachments_returns_container_blobs(mock_container_client):
...
+    job.upload_input_data = Mock(return_value="uploaded-uri")
+    container_client = mock_container_client.from_container_url.return_value
+    container_client.list_blobs.return_value = []
+    container_client.get_blob_client.return_value.download_blob.return_value.readall.return_value = b"data"
+
+    job.upload_attachment("upload", b"data")
+    attachments = job.list_attachments()
+    downloaded = job.download_attachment("download")
+
+    workspace.get_container_uri.assert_has_calls(
```
rigidit: did you consider to cache get_sas_uri call? and we can re-validate it with the same expiry/permission check

rigidit (changes requested):

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: Addressed in most recent commit. _get_attachment_container_uri() now accepts Literal["r", "w", "l"] instead of an unrestricted str and retains an explicit str return type

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: Addressed in most recent commit. Malformed st or se timestamps now produce a debug-level message and cause the URI to be treated as unusable, so a fresh SAS is requested. The log does not include the SAS URI or credentials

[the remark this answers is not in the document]
v-elegacheva: Addressed in most recent commit. The validator now parses the optional SAS st parameter and reuses the uri only after its start time. Future-start tokens are treated as unusable and refreshed. Added regression coverage

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: So yes, this branch is not used by normal current Prod submission flows bc both DPs return an unsigned stored container uri. It remains reachable only for unsubmitted/locallyconstructed jobs and legacy responses. I kept it for compatibility but if those scenarios are not intended to be supported, I can simplify the implementation to always use the cached/refreshed workspace SAS instead

azure-quantum/tests/test_job_attachments.py
v-elegacheva: Addressed in most recent commit with a private per-job cache for freshly requested attachment container SAS URIs. Cached values are revalidated for HTTPs, required permission, start time, and expiry before every reuse. The cache is associated with the current container hostname/pth and is invalidated when job details change including in-place container_uri changes. Explicit method-level URI overrides continue to bypass the cache

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +399,122 @@ def list_attachments(self) -> list[BlobProperties]:
...
+            return (None, f"/job-{self.id}")
+
+        parsed_uri = urlparse(container_uri)
+        return (parsed_uri.hostname, parsed_uri.path.rstrip("/"))
+
+
+    def _is_attachment_container_uri_usable(
+        self,
+        container_uri: str,
+        required_permission: Literal["r", "w", "l"],
```
rigidit: and we probably can drop that required_permission entirely as we cannot request specific permission from backend anyway?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +399,122 @@ def list_attachments(self) -> list[BlobProperties]:
...
+    ) -> bool:
+
+        parsed_uri = urlparse(container_uri)
+        if parsed_uri.scheme.lower() != "https":
+            return False
```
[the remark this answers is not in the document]
v-elegacheva: addressed in recent commit. The shared SAS validator now requires HTTPS, a hostname, and a nonempty container path before considering a URI usable. The pathless uri regression test uses an unexpired signed sas to exercise this exact branch

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +399,122 @@ def list_attachments(self) -> list[BlobProperties]:
...
+                token_start_time = datetime.fromisoformat(
+                    token_start_query_param[0].replace("Z", "+00:00")
+                )
+                if token_start_time.tzinfo is None:
+                    token_start_time = token_start_time.replace(tzinfo=timezone.utc)
+
+            current_utc_time = datetime.now(tz=timezone.utc)
+            has_started = token_start_time is None or token_start_time <= current_utc_time
+            return has_started and current_utc_time < token_expire_time - timedelta(minutes=5)
+        except ValueError:
```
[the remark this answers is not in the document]
v-elegacheva: Addressed in recent commit. The validator now compares current_time + five-minute buffer < expiry instead of subtracting the buffer from the externally supplied expiry. This avoids underflow for boundary values such as year 1

azure-quantum/tests/test_job_attachments.py
```diff
@@ -34,21 +58,392 @@ def test_list_attachments_returns_container_blobs(mock_container_client):
...
+
+@pytest.mark.parametrize(
+    "cached_container_uri",
+    [EXPIRED_CONTAINER_URI, FUTURE_START_CONTAINER_URI],
+)
+def test_unusable_cached_uri_is_refreshed(cached_container_uri):
+    workspace = Mock()
+    workspace.get_container_uri.return_value = SIGNED_CONTAINER_URI
+    job = _job_with_container(workspace=workspace)
+    job._attachment_container_uri_cache = cached_container_uri
```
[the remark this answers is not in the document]
v-elegacheva: Addressed in recent commit. The parametrized cache test now sets _attachment_container_uri_cache_identity to match the current job container identity, ensuring it reaches sas start/expiry validation rather than passing through identity invalidation

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: agreed, and addressed. I removed required_permission and the operation-specific permission branching. Attachment operations now request one workspace sas, cache it per job, and revalidate HTTPS, container identity, signature, permission presence, start time, and expiry before reuse

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +399,108 @@ def list_attachments(self) -> list[BlobProperties]:
...
-            container_uri = self.workspace.get_container_uri(job_id=self.id)
-        else:
-            container_uri = self._details.container_uri
+        container_uri = self._get_attachment_container_uri()
 
         container_client = ContainerClient.from_container_url(container_uri)
         return list(container_client.list_blobs())
 
 
+    def _get_attachment_container_uri(self) -> str:
```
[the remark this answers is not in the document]
v-elegacheva: Addressed in recent commit. I kept the parameter-less helper bc Workspace.get_container_uri() cannot request operation specific permissions and the backend returns a fixed container sas permission set.
The validator now requires that every cached or refreshed workspace sas contain all attachment capabilities needed across operations r,w, and l. 
Tokens with incomplete permissions are refreshed or rejected with focused coverage for both cached and newly refreshed sas URIs

v-elegacheva:
> ## Copilot review overview
> ### 🔵 Needs a closer look
> A valid job-level SAS is unnecessarily refreshed instead of reused.
> 
> **Review effort:** Balanced **Findings:** None
> 
> **Resolved since last review (1)**
> * Medium severity [Require operation-specific SAS permissions for attachment URI reuse](#discussion_r4076431531)
> 
> **Previously missed (1)**
> In code that hasn't changed since last review
> 
> Medium severity Reuse usable stored SAS URI before refreshing container URI

This refresh is intentional. Current Dpv1 and Dpv2 Production flows store unsigned job container URIs and require workspace-linked storage, so JobDetails.container_uri is used only to identify the expected linked container and storage hostname. 
The first attachment operation requests a workspace-issued SAS, and subsequent operations reuse the validated per-job cache. Caller-supplied external job containers are outside the supported scope, while explicit method-level uri overrides remain unchanged

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +401,115 @@ def list_attachments(self) -> list[BlobProperties]:
...
+        ):
+            return False
+
+        query_params = parse_qs(parsed_uri.query)
+        token_expire_query_param = query_params.get("se")
+        token_start_query_param = query_params.get("st")
+        token_permissions = set(query_params.get("sp", [""])[0])
+        if (
+            not query_params.get("sig")
+            or not _ATTACHMENT_CONTAINER_SAS_PERMISSIONS.issubset(token_permissions)
```
rigidit: This is the inverse of what I suggested. My argument was that permission checking is unproductive — the SDK can't request permissions (BlobDetails has only containerName/blobName), so a refresh can never remedy a permission gap. The conclusion was to drop the check and let storage answer authoritatively. Instead the check became a stricter, global precondition that fails harder. test_upload_attachment_rejects_refreshed_uri_without_full_attachment_permissions now pins rl-rejection as intended behavior.

Suggestion: validate only what a refresh can actually fix — sig present and not near expiry (the st handling is a genuinely good addition, keep it). Let a permission shortfall surface as the storage service's own 403, which names the real cause.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +401,115 @@ def list_attachments(self) -> list[BlobProperties]:
...
+        container_uri = self._get_attachment_container_uri()
 
         container_client = ContainerClient.from_container_url(container_uri)
         return list(container_client.list_blobs())
 
 
+    def _get_attachment_container_uri(self) -> str:
+        """Return a validated workspace-issued SAS URI for the job's attachment container.
+
+        The first call refreshes the unsigned URI stored in job details. Later calls reuse the
```
rigidit: we shouldn't set container sas uri to job details right?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +401,115 @@ def list_attachments(self) -> list[BlobProperties]:
...
+        self._attachment_container_uri_cache_identity = container_identity
+        return refreshed_container_uri
+
+
+    def _get_attachment_container_identity(
+        self,
+        container_uri: Optional[str],
+    ) -> tuple[Optional[str], str]:
+        if container_uri is None:
+            return (None, f"/job-{self.id}")
```
rigidit: _get_attachment_container_identity hardcodes f"/job-{self.id}", which is a third copy of the job-{id} convention (also in container_name and workspace.get_container_uri's container_name_format)

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +401,115 @@ def list_attachments(self) -> list[BlobProperties]:
...
+                raise ValueError(
+                    "Refreshed attachment container hostname "
+                    f"'{refreshed_hostname}' does not match job container hostname "
+                    f"'{stored_hostname}'."
+                )
+
+        if not self._is_attachment_container_uri_usable(refreshed_container_uri):
+            raise ValueError(
+                "Refreshed attachment container URI does not contain a usable SAS token."
+            )
```
rigidit: we can skip this check as we assume that backend always return usable sas uri

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +401,115 @@ def list_attachments(self) -> list[BlobProperties]:
...
+                container_name=self.container_name,
+            )
+            stored_hostname = urlparse(container_uri).hostname
+            refreshed_hostname = urlparse(refreshed_container_uri).hostname
+            if stored_hostname != refreshed_hostname:
+                raise ValueError(
+                    "Refreshed attachment container hostname "
+                    f"'{refreshed_hostname}' does not match job container hostname "
+                    f"'{stored_hostname}'."
+                )
```
rigidit: do we need to compare hostnames here? we are operating with workspace with linked storage here and it won't change for the workspace lifecycle right?

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +401,115 @@ def list_attachments(self) -> list[BlobProperties]:
...
+                    token_start_time = token_start_time.replace(tzinfo=timezone.utc)
+
+            current_utc_time = datetime.now(tz=timezone.utc)
+            has_started = token_start_time is None or token_start_time <= current_utc_time
+            return has_started and current_utc_time + timedelta(minutes=5) < token_expire_time
+        except ValueError:
+            logger.debug(
+                "Unable to parse attachment SAS start or expiry time; requesting a fresh URI."
+            )
+            return False
```
rigidit: narrow this method to the one condition a refresh can genuinely fix — expiry — and let storage adjudicate the rest.

rigidit (changes requested):

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: addressed. I removed the permission checking entirely + sig, HTTPS, hostname and path validation. The cache now checks only SAS expiry. A permission shortfall now surfaces as storage's own error instead of sdk precondition

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: I removed _get_attachment_container_identity() completely. The cache is now keyed on the existing container_name property, so there is no separate job-{id} formatting duplicated for the cache

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: Okay! The check has been removed

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: I removed that. I confirmed with backend that dpv2 production and all sas-issuing endpoints validate the account name against the workspace's linked storage account server-side, so this check was indeed redundant for the linked-storage flows

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: Done! The cache reuse check (renamed to _is_attachment_container_uri_unexpired) now validated only se expiry with the 5 min buffer. 
sig, st and permissions are not checked anymore

azure-quantum/azure/quantum/job/base_job.py
v-elegacheva: Indeed so. I checked again- the refreshed sas is only stored in a private per-job cache and is never written back to job.details.container_uri. Updated in recent commit

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +399,69 @@ def list_attachments(self) -> list[BlobProperties]:
...
+
+        Workspace-issued URIs are cached privately on the job and are never written back to job
+        details. A cached URI is reused while its container name is unchanged and its SAS expiry
+        remains outside the renewal window.
+        """
+        container_uri = self._details.container_uri
+        container_name = self.container_name
+        cached_container_uri = self._attachment_container_uri_cache
+        if cached_container_uri:
+            if self._attachment_container_uri_cache_container_name != container_name:
```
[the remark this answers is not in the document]
v-elegacheva: So I checked Workspace.get_container_uri(). Its signature is (job_id=None, container_name=None, container_name_format="job-{job_id}"). It never reads job.details.container_uri or any hostname. The account is fixed by self.storage on the Workspace instance itself.

I confirmd this with an executed test (test_cache_reuse_after_hostname_mutation_never_queries_by_hostname). 
It mutates job.details.container_uri to a different storage account host while keeping the same container name. Then asserts 
workspace.get_container_uri.assert_called_once_with(job_id= .... , container_name= ....) 
to make sure no hostname argument ever crosses that boundary and that both operations resolve to the identical workspace-issued URI. 
So the cache being keyed on container_name alone does not reuse a "different account's" sas in this scenario. It is the same one real account's sas both times, since get_container_uri() structurally cannot fetch based on hostname. I do not think this reachable through this code path, so I am leaving the cache key as is.

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +399,69 @@ def list_attachments(self) -> list[BlobProperties]:
...
+            refreshed_container_uri = self.workspace.get_container_uri(job_id=self.id)
+        else:
+            refreshed_container_uri = self.workspace.get_container_uri(
+                job_id=self.id,
+                container_name=container_name,
+            )
+
+        self._attachment_container_uri_cache = refreshed_container_uri
+        self._attachment_container_uri_cache_container_name = container_name
+        return refreshed_container_uri
```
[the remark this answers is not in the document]
v-elegacheva: Wrong-account/ HTTP: Not reachable through this code path. Workspace.get_container_uri()'s signiture is (job-id,contianer_name, container_name_format). It never reads a hostname and resolves to exactlyone account (self.storage) on the Workspace instance), confimed by test_cache_reuse_after_hostname_mutation_never_queries_by_hostname. Backend-issued sas URIs are also always built from BlobServiceClient / BlobContainerClient tied to the account's https:// blob endpoint so an http:// response isnt something the backend can structurally return for this flow.

Unsigned / insufficient-permission accepted: Not an oversight. Per feedback - agreed to narrow validation to the one condition a refresh can fix (expiry) and defer signature/permission enforcement to Blob Storage's own authorative response. Backend-issued tokens are always signed with full permissions in practice, so this only affects synthetic/malformed inputs, which several tests exercise to document the pass-through behavior

azure-quantum/azure/quantum/job/base_job.py
```diff
@@ -399,16 +399,69 @@ def list_attachments(self) -> list[BlobProperties]:
...
+
+
+    def _is_attachment_container_uri_unexpired(
```
rigidit: _is_attachment_container_uri_unexpired and _get_blob_uri_with_sas_token still both parse se with the same Z→+00:00 workaround and the same 5-minute margin, should we extract some helper to reuse the logic?

